PR #1, opened 2014-12-31
new tests first commit

orius123:
this is my changes for the new tests so far.

it doesn't work well on my machine...

the main advantage in this change is that it gives the path (task or executable) run data
i.e. inputs outputs....

WDYT?


orius123:
another comment
@meshipeer @meirwah

PR #7, opened 2015-01-04
Add new API for pre-compilation of slang files to produce meta-data

oritstone:
@einatneu @varelasavela @ziso @orius123 @meshipeer 

We need to write a new API for supplying meta-data regarding Slang files before the actual compilation.

Currently there is a need to retrieve for each Slang file the following data:
- inputs (in case of flow or operation)
- list of dependencies conatining for each one the full name + type (executable/sys variable)
- namespace


oritstone:
@orius123 please do a code review on the last commit before merging to the master

Issue #4, opened 2015-01-02
better packaging structure for slang CLI is needed 


meirwah:
currently the batch file is under /appassembler/bin

need to change to maybe:
/slang/bin

meirwah:
@orius123  don't froget to update the web-site too...

meirwah:
@orius123 ? WDYT?

orius123:
looks good, but we still need to zip it. another issue?

meirwah:
we can merge this change... and keep commiting to this branch no?

orius123:
yup. just don't close the issue. 
reviewd :white_check_mark:

meshipeer:
add content library

meirwah:
adding @hararzafrir

meirwah:
@hararzafrir  please colse this when done..

meirwah:
isn't this fixed already @oritstone @hararzafrir ?

oritstone:
@orius123 is it done?

orius123:
no. I need to handle it as part of the build task. do we want ot use this one or to open a new one?

orius123:
ok.
I want to redefine this issue.
@openscore/owners please take two minutes to review :pray: 

### How to build:
- Use *nix OS (it might work on Windows :trollface: )
- Clone score-language repo
- Run:  
```bash
      cd score-language && 
      mvn clean install
```
- Download a stable version of slang-cli and run:
```bash
      slang run build/build_cli_flow.sl
```

## Tasks so the build will work: :wink: 
- [ ] Create a `build` directory in score-language repo
- [ ] Create a flow in `build` directory that:
  - [ ] Clone slang-content repo into `build`.
  - [ ] Runs verifier of the current maven build 
        (from `./score-lang-content-verifier/target/score-lang-content-verifier-<version>.jar`) 
         on the content repo, and checks that it succeeded.
  - [ ] Copy `score-lang-cli/target/slang` to `build` as `slang-cli`
  - [ ] **optional** pre-compile and repack `jython-standalone.jar` 
  - [ ] From build/slang-content repo copy `content` and `python-lib` directories into `build/slang-cli`
  - [ ] Run `pip install -t build/slang-cli/python-lib -r build/slang-cli/python-lib/requirements.txt`
  - [ ] Add execute permissions for `build/slang-cli/bin/slang`
  - [ ] Create build/slang-cli.zip from the slang-cli directory
  - [ ] Create build/slang-cli.tar.gz from the slang-cli directory

[...]

PR #17, opened 2015-01-07
change the "appasmbler" dir name to "slang"

meirwah:
part of #4 soultion, @orius123 WDYT?


meirwah:
of course we need also to automate our packging(including tests+content)

orius123:
Better using maven`s build.directory property.
But if it works it's great

meirwah:
i'm putting this on you @orius123  since it must be fixed with the website change ....

PR #19, opened 2015-01-09
Mysql health check

Bonczidai:
added flow that checks the MySQL server status in a docker container and sends an email with the result

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -162,7 +162,7 @@
             <dependency>
                 <groupId>io.openscore.content</groupId>
                 <artifactId>score-ssh</artifactId>
-                <version>0.0.11</version>
+                <version>0.0.13</version>
             </dependency>
             <dependency>
                 <groupId>io.openscore.content</groupId>
--- a/score-language-tests/src/test/resources/yaml/docker-demo/linux_operations.yaml
+++ b/score-language-tests/src/test/resources/yaml/docker-demo/linux_operations.yaml
@@ -28,7 +28,7 @@ operations:
           methodName: runSshShellCommand
       outputs:
         - response: STDOUT
-        - diskSpace:
+        - diskSpace
       results:
         - SUCCESS
         - FAILURE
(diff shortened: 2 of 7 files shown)


Bonczidai:
Is there any way to split a long input expression (like in the report_mysql_status.yaml for the email body - line 79) into several lines? @meirwah @oritstone @orius123

meirwah:
did you tried to use the ">" sign ?

Bonczidai:
Thanks, it works with ">" ..

meirwah:
in flow "get_mysql_status.yaml" it will be better to do in the outputs section:
+          - mysql_statues: returnResult.replace(':', ' ').split('  ')
+          - uptime: "mysql_statues[1]"
+          - threads: "mysql_statues[3]"
+          - questions: "mysql_statues[5]"
 ....

it will be both more effiecent , and make flow more readable...

meirwah:
please review @oritstone

meirwah:
+1

oritstone:
I have a few comments:
1. @meirwah @orius123 What did we decide about the extension of the files? Shouldn't they all be .sl files?
- In almost all of the new yaml files, the inputs names in the comments are not the same as the flow/operation inputs. They should be the same.
2. Many of the inputs of the SSH operations are missing from the comments. I guess this is because they have default values in the operation. This shouldn't be the case. All the inputs of the operation should appear in the comment, and have description. We can consider adding the default values to the action, or we go with the suggestion in the next bullet.
3. I suggest that we will have a general SSH operation that will reside in the technologies\SSH folder of the content. The operations added here should be changed to flows that have 1 step that calls the SSH operation or step in a flow, and give the SSH operation only the relevant inputs (the commands) and get from it the general outputs (returnResult, returnCode...) and parse them. This way we will have a general purpose SSH operation, and specific usages for the docker use-case. In this case we will have all the inputs of the operation documented, but when using it, we can pass to it only the required ones that don't have default values.

Bonczidai:
@meirwah I am not sure we can do this:
- mysql_statues: returnResult.replace(':', ' ').split(' ')
- uptime: "mysql_statues[1]"
- threads: "mysql_statues[3]"
- questions: "mysql_statues[5]" ....

all the outputs of an operation must evaluate to String (OutputsBinding class bindOutputs method), since "returnResult.replace(':', ' ').split(' ')" evaluates to an array the CLI throws this exception:
"Slang Error : The output expression returnResult.replace(':', ' ').split('  ') does not return String value"

meirwah:
your'e right.. maybe it will be cleaner using linux cmd of 'awk'...

PR #20, opened 2015-01-09
Diskspace health check

TudorLesanHP:
Added diskspace_health_check_flow.yaml + small changes.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -162,7 +162,7 @@
             <dependency>
                 <groupId>io.openscore.content</groupId>
                 <artifactId>score-ssh</artifactId>
-                <version>0.0.11</version>
+                <version>0.0.13</version>
             </dependency>
             <dependency>
                 <groupId>io.openscore.content</groupId>
--- a/score-lang-cli/src/main/java/org/openscore/lang/cli/SlangCLI.java
+++ b/score-lang-cli/src/main/java/org/openscore/lang/cli/SlangCLI.java
@@ -96,9 +96,9 @@ public class SlangCLI implements CommandMarker {
     public List<String> getFlowInputs(
             @CliOption(key = {"", "f", "file"}, mandatory = true, help = "Path to filename. e.g. slang inputs --f C:\\Slang\\flow.yaml") final File file,
             @CliOption(key = {"cp", "classpath"}, mandatory = false,
-                    help = "Classpath , a directory comma separated list to flow dependencies, by default it will take flow file dir") final String classPath)
+                    help = "Classpath , a directory comma separated list to flow dependencies, by default it will take flow file dir") final List<String> classPath)
             throws IOException {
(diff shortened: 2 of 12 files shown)


meirwah:
1. why you added to rest_operations.yaml the namespace "lists.manipulation"?
2. check_availability should be a seprate file... at at has nothing to do with linux_ops, it should be a generic op of "compare int" or something likethat
3. did you check "df -Pkh | grep -v "Filesystem" | awk \'NR==1{print $5}\'" with other linux machine (not ubuntu) ?
4. The change in docker_operations.yaml, is wrong, you will get only dangling images frrom it, but that not what we want, we need one step to get all images and a DIFFERENT step to get ONLY dangling images  . Please revert your change in this file, ADD a new operation with the cmd docker images -f dangling=true -q , and cuild a subflow from it
5. typo in diskspace_health_check_flow.yaml , in the flow name : diskspace_heath_check_flow
6. we need another flow output of how many docker images were actually deleted..
7. you need to resolve conflicts...

TudorLesanHP:
1. it was a mistake, changing back.
2.true, I was not sure where to put that though. I'll do that
3.no, I'll check.
4.5 6 7 - taking care of them

meirwah:
@TudorLesanHP can you fix merge conflicts ?

pom.xml
```diff
@@ -162,7 +162,7 @@
             <dependency>
                 <groupId>io.openscore.content</groupId>
                 <artifactId>score-ssh</artifactId>
-                <version>0.0.11</version>
+                <version>0.0.13</version>
```
meirwah: is this really needed??? why?

score-lang-cli/src/main/java/org/openscore/lang/cli/SlangCLI.java
```diff
@@ -96,9 +96,9 @@ public String setEnvVar(
     public List<String> getFlowInputs(
             @CliOption(key = {"", "f", "file"}, mandatory = true, help = "Path to filename. e.g. slang inputs --f C:\\Slang\\flow.yaml") final File file,
             @CliOption(key = {"cp", "classpath"}, mandatory = false,
-                    help = "Classpath , a directory comma separated list to flow dependencies, by default it will take flow file dir") final String classPath)
+                    help = "Classpath , a directory comma separated list to flow dependencies, by default it will take flow file dir") final List<String> classPath)
```
meirwah: i think you have here commits from other merge...

pom.xml
TudorLesanHP: I don't think it's really needed. You told me at one point to test the new SSH from content and make sure it still works with our flows and I committed the change.

meirwah: ok...

meirwah:
why did you merge this? isn't it the same as the one in content?

Issue #22, opened 2015-01-10
Need to update group.id to "io.openscore"


meirwah:
After the change,
we need to upload the artifact to maven central @hararzafrir  , and then update the website @orius123

meirwah:
@hararzafrir can you upload the last artifact(of score-language) to maven central?

meirwah:
So i understand from @hararzafrir  it is now in Mvaen Central, we just need to verify iy works, and update the website, so i'm moving this to @orius123 ...

orius123:
you changed only group id, not packages, right?

meirwah:
right

PR #23, opened 2015-01-11
change group-id from :

meirwah:
<groupId>org.openscore.lang</groupId>
 TO
 <groupId>io.openscore.lang</groupId>


meirwah:
fix #22

PR #35, opened 2015-01-19
System properties

avimora:


meirwah:
"systemProperties" in the yaml file? is this our convention?

avimora:
Well... that's one of the reasons I didn't like the term... don't like it either...
Imposed by the yaml parser, it has to be identical to the java member name...
Suggestions?

orius123:
Didn't we decided something after all of the mails about it? was it systemProperties??
I don't think that this is a good functional name, and if the yamlParser is the problem we should handle it somehow.

meshipeer:
we agree on system properties
and as i comment to the FS: i think it should be system_properties in the DSL

avimora:
So we go with `system_properties` and `system_property`? should I try to figure a workaround for the parser?

avimora:
@meirwah, @orius123, @meshipeer, @oritstone can you please review & merge?

PR #36, opened 2015-01-19
retrieve dynamically the score version in cli

### LINKED ISSUE #16: score version is hard coded in the CLI
look at class SlangCLI param SCORE_VERSION...

Bonczidai:
fixes #16

### DIFF
--- a/score-lang-cli/src/main/java/org/openscore/lang/cli/SlangBanner.java
+++ b/score-lang-cli/src/main/java/org/openscore/lang/cli/SlangBanner.java
@@ -38,13 +38,13 @@ public class SlangBanner extends SlangNamedProvider implements BannerProvider {
 			sb.append("Slang");
 		}
 		sb.append(System.lineSeparator());
-		sb.append(SlangCLI.getVersion());
+		sb.append(getVersion());
 		return sb.toString();
 	}
 
 	@Override
 	public String getVersion() {
-		return SlangCLI.getVersion();
+		return SlangCLI.SCORE_VERSION;
 	}
 
 	@Override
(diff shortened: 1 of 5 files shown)


meirwah:
Please review @avimora

avimora:
The general idea is fine. You can use spring's propertyplaceholder to inject it directly to your bean with zero code.

avimora:
I think slang version is more appropriate

meirwah:
agreed

meirwah:
@Bonczidai please look at spring's propertyplaceholder, maybe this can be a refernce : http://www.mkyong.com/spring/spring-propertyplaceholderconfigurer-example/

avimora:
In context xml you should add `<context:property-placeholder location="application.properties"/>` and you can refer to it by `${score.version}` or whatever the property key is, and inject it directly to the bean. It can be done with xml or annotations (e.g. `@Value("${score.version}")`)

Bonczidai:
updated - slang version is used, please review ..

score-lang-cli/src/main/java/org/openscore/lang/cli/SlangBanner.java
```diff
@@ -38,13 +42,13 @@ public String getBanner() {
...
 		sb.append(System.lineSeparator());
-		sb.append(SlangCLI.getVersion());
+		sb.append(getVersion());
 		return sb.toString();
 	}
 
 	@Override
 	public String getVersion() {
-		return SlangCLI.getVersion();
+		return slangCLI.getSlangVersion();
```
meirwah: why ?? why not add instead :
@Value("${slang.version}")
private String slangVersion;

avimora: Agree

meirwah:
@Bonczidai  can you look at the comments we made here?

Bonczidai:
updated

meirwah:
@avimora  please review...

avimora:
gr8 :100: can I delete the branch after merge?

meirwah:
yes

PR #47, opened 2015-01-23
update complex_navigation test and check_number operation to use string input

Bonczidai:
updated operation and tests accordingly so it can also be run from the CLI


orius123:
do we publish this operation or just for tests?

Bonczidai:
I think it is just for tests in score-language repository;
however I used it for demos because it's a simple example how we can use navigation;
to be able to trigger it from cli: I had to change the check number operation because it expected an int input and from the cli we only get strings

PR #67, opened 2015-01-29
fixed a mistake in unsubscribe method's comment plus other readability changes

SamMarkowitz:
unSubscribeOnEvent method's comment was incorrect.


meirwah:
:+1:

PR #77, opened 2015-02-02
Align operation slang structure to flow

oritstone:
- 1 operation in file
- name of file = name of operation
- starts with operation: attribute, and below a name: attribute


score-lang-api/src/main/java/org/openscore/lang/api/Slang.java
```diff
@@ -26,25 +26,16 @@
 public interface Slang {
 
     /**
-     * Compile a flow written in slang
+     * Compile a flow/operation written in slang
```
orius123: as @SamMarkowitz commented on another API it's preferable to use 'or' and not a slash

score-lang-api/src/main/java/org/openscore/lang/api/Slang.java
```diff
@@ -26,25 +26,16 @@
...
      * @param source the slang source containing the flow
-     * @param dependencies a set of slang sources of of all the flow's dependencies
+     * @param dependencies a set of slang sources of of all the flow/operation's dependencies
```
orius123: same goes here

score-lang-api/src/main/java/org/openscore/lang/api/Slang.java
```diff
@@ -26,25 +26,16 @@
...
-     * @param operationName the name of the operation to compile from the source
-     * @param dependencies a set of slang sources of all the operation's dependencies
-     * @return the compiled artifact of the operation
-     */
-    public CompilationArtifact compileOperation(SlangSource source, String operationName, Set<SlangSource> dependencies);
-
 	/**
 	 * Run a flow or operation written in slang already compiled to a compilationArtifact
-	 * @param compilationArtifact the compiled artifact of the flow
+	 * @param compilationArtifact the compiled artifact of the flow/operation
```
orius123: as @SamMarkowitz commented on another API it's preferable to use 'or' and not a slash

score-lang-compiler/src/main/java/org/openscore/lang/compiler/SlangCompilerImpl.java
```diff
@@ -181,45 +170,32 @@ private ExecutionPlan compileToExecutionPlan(Executable executable) {
...
-            Map.Entry<String, Map<String, Object>> entry = operation.entrySet().iterator().next();
-            String operationName = entry.getKey();
-            Map<String, Object> operationRawData;
-            try {
-                operationRawData = entry.getValue();
-            } catch (ClassCastException ex){
-                throw new RuntimeException("Operation: " + operationName + " syntax is illegal. Below operation name, there should be a map of values.");
-            }
-            executables.add(executableBuilder.transformToExecutable(parsedSlang, operationName, operationRawData));
+    private Executable transformOperation(ParsedSlang parsedSlang) {
```
orius123: I prefer that we will have one method for both flow and operation and than we will switch case only once

PR #78, opened 2015-02-02
rename & invert override to overridable

### LINKED ISSUE #41: Rename the `override` attribute on inputs
May be `force`? or invert it's meaning?

avimora:
fixes #41


score-lang-compiler/src/main/java/org/openscore/lang/compiler/transformers/AbstractInputsTransformer.java
```diff
@@ -54,11 +54,11 @@ private static Input createPropInput(Map.Entry<String, Map<String, Serializable>
 		}
 		boolean required = !props.containsKey(REQUIRED_KEY) || (boolean)props.get(REQUIRED_KEY);// default is required=true
 		boolean encrypted = props.containsKey(ENCRYPTED_KEY) && (boolean)props.get(ENCRYPTED_KEY);
-		boolean override = props.containsKey(OVERRIDE_KEY) && (boolean)props.get(OVERRIDE_KEY);
+		boolean overridable = !props.containsKey(OVERRIDABLE_KEY) || (boolean)props.get(OVERRIDABLE_KEY);
```
oritstone: Condition is wrong - According to the condition above, If the author writes overridable: false, you will treat it as true.
Please add also a unit test for this use case

avimora: overridable: false = !true || false = false
There ARE tests for that

meirwah:
:+1:

Issue #5, opened 2015-01-04
Add prints of flow outputs to the CLI...


meirwah:
Maybe we can add prints of flow outputs  to the CLI prints (like in task name).
so it will be easy to see flow result...

TudorLesanHP:
@meirwah I added flow_outputs_to_cli branch with the update

meirwah:
fixed by #89 ...

PR #89, opened 2015-02-06
Modified ScoreServicesImpl to also show outputs in  CLI

tudorlesan:


score-lang-cli/src/main/java/org/openscore/lang/cli/services/ScoreServicesImpl.java
```diff
@@ -131,6 +131,37 @@ public synchronized void onEvent(ScoreEvent scoreEvent) throws InterruptedExcept
                         printWithColor(Ansi.Color.YELLOW, prefix + taskName);
                     }
                     break;
+                case ScoreLangConstants.EVENT_OUTPUT_END:
```
meirwah: all the logic must be in a diff method..
moreover.. i do not understand what you are doing here, why so much parsing is needed?

meirwah:
can you add an image snapshot on how it looks to run this in the CLI?

TudorLesanHP:
cli_outputs

score-lang-cli/src/main/java/org/openscore/lang/cli/services/ScoreServicesImpl.java
TudorLesanHP: I first make sure I print just the actual outputs not also the published ones. Then get the prefix (nr. of '-'s)
When printing the outputs I truncate the value if the length is greater than 30, or put "(empty)" if output has no value.
I also replace newlines and then print the actual output (key - value)

TudorLesanHP:
fixes #5

orius123:
I think it's better to present only the outputs of the main flow and not all of the children.

meirwah:
agree with Eliya

TudorLesanHP:
@orius123 @meirwah , only main flow outputs are shown now

score-lang-cli/src/main/java/org/openscore/lang/cli/services/ScoreServicesImpl.java
```diff
@@ -124,19 +124,51 @@ public synchronized void onEvent(ScoreEvent scoreEvent) throws InterruptedExcept
...
         }
 
+        private void printOutputs(Map<String, Serializable> data) {
+            if(data.containsKey(LanguageEventData.OUTPUTS) && data.containsKey(LanguageEventData.PATH) && data.get(LanguageEventData.PATH).equals("0/0")) {
+
+                @SuppressWarnings("unchecked") Map<String, String> outputs = (Map<String, String>) data.get(LanguageEventData.OUTPUTS);
+
+                String prefix = getPrefix(data);
+
+                if (outputs != null && !outputs.keySet().isEmpty()) {
```
orius123: Would use apache map utils

score-lang-cli/src/main/java/org/openscore/lang/cli/services/ScoreServicesImpl.java
```diff
@@ -124,19 +124,51 @@ public synchronized void onEvent(ScoreEvent scoreEvent) throws InterruptedExcept
...
+                    printWithColor(Ansi.Color.WHITE, prefix + "Outputs:");
+
+                    for (String key : outputs.keySet()) {
+                        String outputValue = outputs.get(key).replace("\n", " ");
+                        outputs.put(key, StringUtils.abbreviate(outputValue, 0, OUTPUT_VALUE_LIMIT));
+                        if (StringUtils.isEmpty(outputValue)) {
+                            outputs.put(key, "(empty)");
+                        }
+                        printWithColor(Ansi.Color.WHITE, prefix + "  " + key + " = " + outputs.get(key));
+                    }
```
orius123: Do we still need the prefix if we only print the flow outputs?

score-lang-cli/src/main/java/org/openscore/lang/cli/services/ScoreServicesImpl.java
```diff
@@ -124,19 +124,51 @@ public synchronized void onEvent(ScoreEvent scoreEvent) throws InterruptedExcept
...
+
                 case EVENT_EXECUTION_FINISHED :
                     printFinishEvent(data);
                     break;
             }
         }
 
+        private void printOutputs(Map<String, Serializable> data) {
+            if(data.containsKey(LanguageEventData.OUTPUTS) && data.containsKey(LanguageEventData.PATH) && data.get(LanguageEventData.PATH).equals("0/0")) {
+
```
orius123: Please use a constant for the root path

score-lang-cli/src/main/java/org/openscore/lang/cli/services/ScoreServicesImpl.java
TudorLesanHP: no we don't need it anymore. Though I suggest printing like we were to get the prefix so it looks better.

meirwah:
i think we can skip the "Outputs" keyword print, enough is to just print the outputs, WDYT?

TudorLesanHP:
fine by me.

score-lang-cli/src/main/java/org/openscore/lang/cli/services/ScoreServicesImpl.java
```diff
@@ -124,19 +128,48 @@ public synchronized void onEvent(ScoreEvent scoreEvent) throws InterruptedExcept
...
+        private void printOutputs(Map<String, Serializable> data) {
+            if(data.containsKey(LanguageEventData.OUTPUTS) && data.containsKey(LanguageEventData.PATH) && data.get(LanguageEventData.PATH).equals(EXEC_START_PATH)) {
+
+                @SuppressWarnings("unchecked") Map<String, String> outputs = (Map<String, String>) data.get(LanguageEventData.OUTPUTS);
+                Map<String, String> modifiableOutputs = new HashMap<>();
+                modifiableOutputs.putAll(outputs);
+                if (MapUtils.isNotEmpty(outputs)) {
+
+                    for (String key : outputs.keySet()) {
+                        String outputValue = modifiableOutputs.get(key).replace("\n", " ");
```
meirwah: this .replace("\n", " "), is error prone.. what if the value of the output is "n\n" ?
also where are the tests for this?

score-lang-cli/src/main/java/org/openscore/lang/cli/services/ScoreServicesImpl.java
```diff
@@ -124,19 +128,48 @@ public synchronized void onEvent(ScoreEvent scoreEvent) throws InterruptedExcept
...
+                        modifiableOutputs.put(key, StringUtils.abbreviate(outputValue, 0, OUTPUT_VALUE_LIMIT));
+                        if (StringUtils.isEmpty(outputValue)) {
+                            modifiableOutputs.put(key, "(empty)");
```
meirwah: i think if the output is empty, just skip it...

score-lang-cli/src/main/java/org/openscore/lang/cli/services/ScoreServicesImpl.java
```diff
@@ -124,19 +128,48 @@ public synchronized void onEvent(ScoreEvent scoreEvent) throws InterruptedExcept
...
                     printFinishEvent(data);
                     break;
             }
         }
 
+        private void printOutputs(Map<String, Serializable> data) {
+            if(data.containsKey(LanguageEventData.OUTPUTS) && data.containsKey(LanguageEventData.PATH) && data.get(LanguageEventData.PATH).equals(EXEC_START_PATH)) {
+
+                @SuppressWarnings("unchecked") Map<String, String> outputs = (Map<String, String>) data.get(LanguageEventData.OUTPUTS);
+                Map<String, String> modifiableOutputs = new HashMap<>();
```
meirwah: all this (lines 150-152) can be shorten to a simple : 
Map<String, String> outputs = new HashMap((Map<String, String>)data.get(LanguageEventData.OUTPUTS))

score-lang-cli/src/main/java/org/openscore/lang/cli/services/ScoreServicesImpl.java
TudorLesanHP: I can remove this then, but outputs may be oddly shown in the CLI then.

orius123:
@tudorlesan made some cosmetic fixes and merged from master.
what happens in cases where outputs are not strings?

tudorlesan:
I tested with this solution and if a flow output was not String it threw an error. So I changed the Map from <String,String> to <String,Serializable> and everything seems to be working fine. When checking empty values or too long values the toString() is applied.

oritstone:
@tudorlesan is this pull request ready for review & merge?

tudorlesan:
@oritstone yes, it's ready

PR #90, opened 2015-02-06
Added quiet mode for CLI

### LINKED ISSUE #64: Quiet mode for slang-cli 'run' command
'run' currently prints out volumes of extra information that would be very awkward to use as part of a script, or anything else that parses results, eg:

```text
$ ../score-lang-cli/slang/bin/slang run division.sl --i input1=10,input2=-1
Database name is: H2 databaseMajorVersion is: 1
- divider
- printer
10/-1 = -10.0
Flow : division finished with result : SUCCESS
Jan 28, 2015 6:00:22 PM org.springframework.shell.core.AbstractShell handleExecutionResult
INFO: Flow execution time took  0:00:03.158 , with execution id : 101600001
```
A quiet mode, that only returns what is specified as output by the flow would be very useful.  In this case, that would be:
```
10/-1 = -10.0
```
Additionally (and I'll file a separate issue on this), it would be a major plus if the exit code reflected the success or failure of the flow.

tudorlesan:
Fixes #64


score-lang-cli/src/main/java/org/openscore/lang/cli/SlangCLI.java
```diff
@@ -64,7 +64,8 @@ public String run(
             @CliOption(key = {"", "f", "file"}, mandatory = true, help = "Path to filename. e.g. slang run --f C:\\Slang\\flow.yaml") final File file,
             @CliOption(key = {"cp", "classpath"}, mandatory = false, help = "Classpath , a directory comma separated list to flow dependencies, by default it will take flow file dir") final List<String> classPath,
             @CliOption(key = {"i", "inputs"}, mandatory = false, help = "inputs in a key=value comma separated list") final Map<String, Serializable> inputs,
-            @CliOption(key = {"spf", "system-property-file"}, mandatory = false, help = "comma separated list of system property file locations") final List<String> systemPropertyFiles) throws IOException {
+            @CliOption(key = {"spf", "system-property-file"}, mandatory = false, help = "comma separated list of system property file locations") final List<String> systemPropertyFiles,
```
meirwah: what is this related too?

score-lang-cli/src/main/java/org/openscore/lang/cli/services/ScoreServices.java
```diff
@@ -24,4 +24,5 @@
     void subscribe(ScoreEventListener eventHandler, Set<String> eventTypes);
     Long trigger(CompilationArtifact compilationArtifact, Map<String, ? extends Serializable> inputs, Map<String, ? extends Serializable> systemProperties);
     Long triggerSync(CompilationArtifact compilationArtifact, Map<String, ? extends Serializable> inputs, Map<String, ? extends Serializable> systemProperties);
+    Long triggerSync(CompilationArtifact compilationArtifact, Map<String, ? extends Serializable> inputs, Map<String, ? extends Serializable> systemProperties, boolean quiet);
```
meirwah: why new method?

score-lang-cli/src/main/java/org/openscore/lang/cli/services/ScoreServicesImpl.java
```diff
@@ -95,11 +96,47 @@ public Long triggerSync(CompilationArtifact compilationArtifact, Map<String, ? e
         slang.unSubscribeOnEvents(scoreEventListener);
         return executionId;
     }
+    @Override
+    public Long triggerSync(CompilationArtifact compilationArtifact, Map<String, ? extends Serializable> inputs, Map<String, ? extends Serializable> systemProperties, boolean isQuiet){
```
meirwah: this is code duplication.. why not add the param quiet to existing method?

score-lang-cli/src/main/java/org/openscore/lang/cli/SlangCLI.java
TudorLesanHP: don't know why this appeared. Didn't intend to.

score-lang-cli/src/main/java/org/openscore/lang/cli/services/ScoreServices.java
TudorLesanHP: I meant to delete the old one. I didn't have time to check if everything still works fine so I overloaded the method.

score-lang-cli/src/main/java/org/openscore/lang/cli/services/ScoreServicesImpl.java
TudorLesanHP: again, I overloaded the method to make sure everything still works before checking it. I will make sure there only one method

score-lang-cli/src/main/java/org/openscore/lang/cli/SlangCLI.java
meirwah: again.. update from master...

meirwah:
@TudorLesanHP please fix issues and add test before pull request..

meirwah:
@orius123 or @avimora  can you review this too?

TudorLesanHP:
fixes #64

score-lang-cli/src/main/java/org/openscore/lang/cli/SlangCLI.java
```diff
@@ -72,9 +73,15 @@ public String run(
...
+            if(logValue == null || logValue.equals("DEBUG")) {
+                id = scoreServices.triggerSync(compilationArtifact, inputs, systemProperties, false);
+            }else if(logValue.equals("INFO")){
+                id = scoreServices.triggerSync(compilationArtifact, inputs, systemProperties, true);
+            }else{
+                return "Wrong value for --log. Valid: INFO, DEBUG.";
+            }
             stopWatch.stop();
-            return triggerSyncMsg(id, stopWatch.toString());
+            return (logValue == null || logValue.equals("DEBUG")) ? triggerAsyncMsg(id, compilationArtifact.getExecutionPlan().getName()) : "";
```
meirwah: i do not like the complexity here.. there should be only -Q param in my opinion
it is either quiet or not.. let's not make things complex for no reason

score-lang-cli/src/main/java/org/openscore/lang/cli/services/ScoreServicesImpl.java
```diff
@@ -131,12 +143,43 @@ public synchronized void onEvent(ScoreEvent scoreEvent) throws InterruptedExcept
...
                 case EVENT_EXECUTION_FINISHED :
+                   if (!flowFinished.get()){
+                        flowFinished.set(true);
+                    }
                     printFinishEvent(data);
                     break;
             }
         }
 
+        private void printFlowOutputs(Map<String, Serializable> data) {
```
meirwah: why is this so complex???

score-lang-cli/src/main/java/org/openscore/lang/cli/services/ScoreServicesImpl.java
```diff
@@ -131,12 +143,43 @@ public synchronized void onEvent(ScoreEvent scoreEvent) throws InterruptedExcept
                         printWithColor(Ansi.Color.YELLOW, prefix + taskName);
                     }
                     break;
+                case ScoreLangConstants.EVENT_OUTPUT_END:
+                    if(quietListener){
```
meirwah: isn't the the way around?

score-lang-cli/src/test/java/org/openscore/lang/cli/SlangCLITest.java
```diff
@@ -45,9 +45,9 @@
 public class SlangCLITest {
 
     private final static String[] CONTEXT_PATH = { "classpath*:/META-INF/spring/test-spring-shell-plugin.xml" };
-    private final static String FLOW_PATH_SLAH = "C:\\\\flow.yaml";
+    private final static String FLOW_PATH_BACKSLASH_INPUT = "C:\\\\flow.yaml";
```
meirwah: why was this changed?

TudorLesanHP: the variable name was not consistent since the value contained backslash not forwardslash

meirwah: ok

score-lang-cli/src/main/java/org/openscore/lang/cli/services/ScoreServicesImpl.java
TudorLesanHP: the quiet mode refers to not showing tasks that are executed. The flow outputs are printed even if quietListener is true. atm the implementation is: normal mode - print tasks (no outputs), quiet mode - print flow outputs (no tasks)
Once we merge the flow_outputs_to_cli branch the normal mode will print flow outputs by default and the condition won't be necessary.

score-lang-cli/src/main/java/org/openscore/lang/cli/services/ScoreServicesImpl.java
TudorLesanHP: this gets the outputs of the flow, truncates any value to length 30 , puts "(empty)" in empty values and replaces newlines with space. it isn't so complex, what do you think I could simplify/remove?

score-lang-cli/src/main/java/org/openscore/lang/cli/SlangCLI.java
TudorLesanHP: I agree that the -Q option would be simpler, but as far as I know, SpringShell doesn't offer us this posibility

score-lang-cli/src/main/java/org/openscore/lang/cli/services/ScoreServicesImpl.java
meirwah: don't hard code numbrers like 30, use StringUtils to simplify ssome stuff here..

score-lang-cli/src/main/java/org/openscore/lang/cli/SlangCLI.java
meirwah: @CliOption(key = {"", "q", "quiet"}, mandatory = false, help = "quiet", specifiedDefaultValue = "true",unspecifiedDefaultValue = "false") final Boolean quiet

TudorLesanHP:
@meirwah fixed

score-lang-cli/src/main/java/org/openscore/lang/cli/services/ScoreServicesImpl.java
```diff
@@ -106,37 +117,70 @@ public boolean isFlowFinished() {
...
+                        String prefix = getPrefix(data);
                         printWithColor(Ansi.Color.YELLOW, prefix + taskName);
                     }
                     break;
-                case EVENT_EXECUTION_FINISHED :
+                case ScoreLangConstants.EVENT_OUTPUT_END:
+                        printOutputs(data);
+                    break;
+                case EVENT_EXECUTION_FINISHED:
+                    if (!flowFinished.get()) {
```
meirwah: why was this added?
how does it relate to the issue?

score-lang-cli/src/main/java/org/openscore/lang/cli/services/ScoreServicesImpl.java
```diff
@@ -106,37 +117,70 @@ public boolean isFlowFinished() {
...
+                        flowFinished.set(true);
+                    }
                     printFinishEvent(data);
                     break;
             }
         }
 
+        private void printOutputs(Map<String, Serializable> data) {
```
meirwah: why on a fix for quiet mode, you have changes to the outputs print method?

score-lang-cli/src/main/java/org/openscore/lang/cli/services/ScoreServicesImpl.java
TudorLesanHP: "A quiet mode, that only returns what is specified as output by the flow would be very useful". But when I first wrote it, the printOutputs method for the quiet_mode_for_cli branch was different than the printOutputs method for the outputs_in_cli branch since one printed just flow outputs and the other also printed step outputs. Now they are the same, so the code for the method is on both branches.

score-lang-cli/src/main/java/org/openscore/lang/cli/services/ScoreServicesImpl.java
TudorLesanHP: both #5 and #64 required outputs to be printed on the console. I first wrote the printOutputs method for #5 which printed all outputs including step outputs. Then I wrote the printOutputs for #64 which only printed flow outputs. Now since I changed the one on #5 both methods are the same on both branches.

score-lang-cli/src/main/java/org/openscore/lang/cli/SlangCLI.java
meirwah: it is still here, @TudorLesanHP  are you going to update from master?

TudorLesanHP: I already did. No conflict appeared. master doesn't even appear any more on the "Merge branches"

score-lang-cli/src/main/java/org/openscore/lang/cli/services/ScoreServicesImpl.java
meirwah: @TudorLesanHP that is a bad apporach, each branch should hold ONLY changes relevent to what the bug will fix, after you merge one of them, you will go to the other branch, and fix accordingly, what you are doing is not by the Git bast practices...

score-lang-cli/src/main/java/org/openscore/lang/cli/SlangCLI.java
```diff
@@ -72,9 +73,9 @@ public String run(
         if (!triggerAsync) {
             StopWatch stopWatch = new StopWatch();
             stopWatch.start();
-            id = scoreServices.triggerSync(compilationArtifact, inputs, systemProperties);
+            id = scoreServices.triggerSync(compilationArtifact, inputs, systemProperties, quiet);
             stopWatch.stop();
-            return triggerSyncMsg(id, stopWatch.toString());
+            return (!quiet) ? triggerAsyncMsg(id, compilationArtifact.getExecutionPlan().getName()) : "";
```
meirwah: this looks better (less verbose):
quiet ? StringUtils.EMPTY : triggerAsyncMsg(id, compilationArtifact.getExecutionPlan().getName())

meirwah: you changed the method name from triggerSyncMsg to triggerAsyncMsg ? why? it's to diff methods?

score-lang-cli/src/main/java/org/openscore/lang/cli/SlangCLI.java
```diff
@@ -72,9 +73,9 @@ public String run(
...
         }
         id = scoreServices.trigger(compilationArtifact, inputs, systemProperties);
         return triggerAsyncMsg(id, compilationArtifact.getExecutionPlan().getName());
```
meirwah: so the Async method does not change?? why?

score-lang-cli/src/main/java/org/openscore/lang/cli/services/ScoreServicesImpl.java
TudorLesanHP: I know. The changes were unique for each branch. I first did #5, then #64. I understood that each issue should have different outputs being printed. But now they contain the same logic after the changes to #5. After you merge one of the branches. I will go to the other one and fix any conflicts.

score-lang-cli/src/main/java/org/openscore/lang/cli/SlangCLI.java
TudorLesanHP: my mistake. Sorry about that

score-lang-cli/src/main/java/org/openscore/lang/cli/SlangCLI.java
TudorLesanHP: I only added the option for the synchronous execution. You said in #64 that the asynchronous execution doesn't print tasknames, so quiet mode wouldn't affect it, would it?

score-lang-cli/src/main/java/org/openscore/lang/cli/services/ScoreServicesImpl.java
```diff
@@ -69,18 +71,27 @@ public Long trigger(CompilationArtifact compilationArtifact, Map<String, ? exten
...
         //add start event
         Set<String> handlerTypes = new HashSet<>();
-        handlerTypes.add(EventConstants.SCORE_FINISHED_EVENT);
-        handlerTypes.add(EventConstants.SCORE_ERROR_EVENT);
-        handlerTypes.add(EventConstants.SCORE_FAILURE_EVENT);
-        handlerTypes.add(SLANG_EXECUTION_EXCEPTION);
-        handlerTypes.add(EVENT_EXECUTION_FINISHED);
-        handlerTypes.add(EVENT_INPUT_END);
+        if(isQuiet){
+            handlerTypes.add(EVENT_OUTPUT_END);
```
meirwah: why was this added? does it mean that step outputs are always printed?

score-lang-cli/src/main/java/org/openscore/lang/cli/services/ScoreServicesImpl.java
meirwah: please remove the ouputs changes that are unrelated..
this nmakes it impossible to track the logic of the change..

score-lang-cli/src/main/java/org/openscore/lang/cli/services/ScoreServicesImpl.java
TudorLesanHP: This just specifies what events are tracked in both modes. Quiet mode only tracks outputs and the finish event while normal mode also tracks all others. The printOutputs method checks so that only the flow outputs get printed. So step outputs are never printed. Should I also add the exception/error events to be tracked for quiet mode? ATM no errors are printed to CLI in quiet mode.

score-lang-cli/src/main/java/org/openscore/lang/cli/services/ScoreServicesImpl.java
TudorLesanHP: #64 requires flow outputs to be printed (#5). Maybe first merge #5 to master then I can remove the output changes on #64 then I can get them from the master?

meirwah:
@orius123 please review this too...

meirwah:
more tests are needed here.. 
where are tasts for the async?

meirwah:
@tudorlesan please update this branch...
@orius123 / @avimora  can you review this too?

meirwah:
@avimora can you review this too?

meirwah:
@tudorlesan please update branch...

score-lang-cli/src/main/java/org/openscore/lang/cli/SlangCLI.java
```diff
@@ -85,12 +86,12 @@ public String run(
         if (!triggerAsync) {
             StopWatch stopWatch = new StopWatch();
             stopWatch.start();
-            id = scoreServices.triggerSync(compilationArtifact, mergedInputs, systemProperties);
+            id = scoreServices.triggerSync(compilationArtifact, mergedInputs, systemProperties, quiet);
             stopWatch.stop();
-            return triggerSyncMsg(id, stopWatch.toString());
+            return quiet ? StringUtils.EMPTY : triggerSyncMsg(id, compilationArtifact.getExecutionPlan().getName());
         }
```
meirwah: compilationArtifact.getExecutionPlan().getName() for the duration param????
that does not make any sense....

tudorlesan: sorry, do not know how that got in there. will fix it asap

meirwah: Please add test to verify it works...

tudorlesan: Fixed this. I already added tests for quiet mode. This just printed the flow name instead of the execution duration, but I can't really test this since the duration could be different for different systems.

score-lang-cli/src/main/java/org/openscore/lang/cli/services/SyncTriggerEventListener.java
```diff
@@ -93,6 +93,7 @@ public synchronized void onEvent(ScoreEvent scoreEvent) throws InterruptedExcept
                 break;
 
             case EVENT_EXECUTION_FINISHED :
+                flowFinished.set(true);
```
meirwah: still not sure it is really needed... @orius123  WDYT?

orius123: On the contrary, I think we should stop listening to score finished event. And only to slang event.

meirwah:
so :+1: to merge this...

meirwah:
fixes #64

score-lang-cli/src/main/java/org/openscore/lang/cli/SlangCLI.java
```diff
@@ -85,12 +86,12 @@ public String run(
...
             stopWatch.start();
-            id = scoreServices.triggerSync(compilationArtifact, mergedInputs, systemProperties);
+            id = scoreServices.triggerSync(compilationArtifact, mergedInputs, systemProperties, quiet);
             stopWatch.stop();
-            return triggerSyncMsg(id, stopWatch.toString());
+            return quiet ? StringUtils.EMPTY : triggerSyncMsg(id, stopWatch.toString());
         }
-        id = scoreServices.trigger(compilationArtifact, mergedInputs, systemProperties);
-        return triggerAsyncMsg(id, compilationArtifact.getExecutionPlan().getName());
+        id = scoreServices.trigger(compilationArtifact, inputs, systemProperties);
```
tudorlesan: I think here it was mergedInputs instead of inputs. Fixed this in cli_extentions_support

Issue #97, opened 2015-02-10
CLI input value cannot contain comma(,)


TudorLesanHP:
I wanted to run the http_client_action operation in the CLI. The body input was JSON and should have contained commas, but that is the delimiter for the inputs.

TudorLesanHP:
I propose using "\," for the input value. The MapConverter can firstly replace "\," with a different char/sequence of chars (like "&|" or whatever characters have less chance of appearing in the CLI) 

Then as usual the string value gets split into the different inputs with delimiter "," and afterwards that char/seq.char from each input value can be replaced back with ",". 

WDYT @meirwah @orius123

Bonczidai:
Another suggestion:

- use syntax: --i `"normal_input=value,json_input='{\"id\": 1,\"name\": \"Foo\",\"price\": [1,2,3]}'"` 
- the MapConverter gets a string like this: `normal_input=value,json_input='{"id": 1,"name": "Foo","price": [1,2,3]}'`
- two cases in the converter:
    - normal: value
    - special: 'between marks' - take the value between marks unmodified
- note: special charachters like `"` need to be escaped, others like `,` are ok

orius123:
Can we separate inputs with space instead of a comma?

Bonczidai:
I think it is possible to separate inputs with space if we put the inputs between `"..."` and change of course the converter ... but that does not solve the json issue (since it contains not only commas but spaces and other characters as well)

meirwah:
maybe give in the CLI ability to load inputs from file? like System Prop ? will solve this?

Bonczidai:
:+1:

orius123:
I think that passing json as string is escaping hell anyway, can we parse the inputs ourselves without wrapping with "" ?

TudorLesanHP:
@Bonczidai you don't need to escape quotes(") in the json_input if you put it in the CLI. Also , with your solution, what if we really want the input value to contain single quotes (')

TudorLesanHP:
@meirwah loading inputs from file will definitely solve this but I was thinking that it may be a good enhancement since we may need inputs that contain commas in the CLI.

TudorLesanHP:
@orius123 the MapConverter just gets a string(all the inputs) and from then on we can delimit/parse it however we want.

orius123:
if comma is a special character in out app we need to provide a way to escape it. File is another thing.

avimora:
Fixed by PR #99

PR #99, opened 2015-02-11
Changed MapConverter so that commas can be used as input values.

TudorLesanHP:
Now just escape the comma in the value. 
Example: run --f filepath --i input1=firstValue\,secondValue,input2=thirdValue. 

Notice the second comma is not escaped so everything after will be split into input2

### DIFF
--- a/score-lang-cli/src/main/java/org/openscore/lang/cli/converters/MapConverter.java
+++ b/score-lang-cli/src/main/java/org/openscore/lang/cli/converters/MapConverter.java
@@ -30,6 +30,9 @@ import java.util.Map;
 
 @Component
 public class MapConverter implements Converter<Map<String, String>> {
+
+    public static final String ESCAPE_EXPRESSION = "\\&^\\&";
+
     @Override
     public boolean supports(Class<?> type, String optionContext) {
         return Map.class.isAssignableFrom(type);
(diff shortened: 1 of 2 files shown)


TudorLesanHP:
fixes #97

score-lang-cli/src/main/java/org/openscore/lang/cli/converters/MapConverter.java
```diff
@@ -37,12 +37,14 @@ public boolean supports(Class<?> type, String optionContext) {
...
     @Override
     public Map<String, String> convertFromText(String value, Class<?> targetType, String optionContext) {
+        value = value.replace("\\,", "\\&^\\&");
         String[] values = StringUtils.commaDelimitedListToStringArray(value);
         Map<String, String> map = new HashMap<>();
 
         for (String v : values) {
             String[] keyValue = StringUtils.delimitedListToStringArray(v, "=");
             if (keyValue.length == 2) {
+                keyValue[1] = keyValue[1].replace("\\&^\\&", ",");
```
orius123: Please extract the escape expression to a variable

meirwah:
Please add tests for this..

SamMarkowitz:
Added to the docs in openscore.github.io/3d86558a78903ca4816da7a0dfe545df54dbd6bc.

PR #100, opened 2015-02-11
modify the exit code of spring-shell in case of slang errors

### LINKED ISSUE #65: slang-cli exit code is success when flow fails
slang-cli returns an exit code of 0, independent of the success or failure of the flow that has been invoked.  This makes it very annoying to try and include calls to run a flow inside of other scripting, since custom logic is then needed to determine if the flow succeeded or failed.

The flow runs successfully:
```
$ ../score-lang-cli/slang/bin/slang run division.sl --i input1=10,input2=-1
Database name is: H2 databaseMajorVersion is: 1
- divider
- printer
10/-1 = -10.0
Flow : division finished with result : SUCCESS
Jan 28, 2015 6:00:22 PM org.springframework.shell.core.AbstractShell handleExecutionResult
INFO: Flow execution time took  0:00:03.158 , with execution id : 101600001
$ echo $?
0
```

The flow fails:

```
$ ../score-lang-cli/slang/bin/slang run division.sl --i input1=10,input2=a 
Database name is: H2 databaseMajorVersion is: 1
- divider
Slang Error : Error executing python script: Traceback (most recent call last):
  File "<string>", line 4, in <module>
ValueError: invalid literal for __float__: a

Jan 28, 2015 6:10:00 PM org.springframework.shell.core.AbstractShell handleExecutionResult
INFO: Flow execution time took  0:00:02.913 , with execution id : 101600001
$ echo $?
0
```

Bonczidai:
In case of slang errors throw exception for spring-shell to modify the exit code of the application - fixes #65 


Bonczidai:
success case:
success_case

failure case:
exit_slang_error

meirwah:
i think the colors have changed ..
the duration line does not have color now ...

@orius123  @avimora please review...

meirwah:
also the ErrorEvent msg does not seem to have color anymore...
and where is the tests for this?

Bonczidai:
the messages does not have color because the flows in the previous screenshots were not run in interactive mode (they were run like "slang run --f ...") - in interactive mode they look like this:
success_case2

However in error case the duration line is not printed now:
before:
error_message_1

after:
error_message_2

Do we want that line printed in error case?

meirwah:
i think we can live with this..
@orius123 @meshipeer @oritstone @avimora WDYT?

Bonczidai:
@avimora Could you please review this PR?

PR #101, opened 2015-02-11
Add verifier tool for slang content

oritstone:


oritstone:
@orius123 please review and merge if OK
Still missing tests & static verification, so leave branch

score-lang-content-verifier/src/main/java/org/openscore/lang/tools/verifier/SlangFilesVerifier.java
```diff
@@ -0,0 +1,35 @@
...
+import org.springframework.context.annotation.AnnotationConfigApplicationContext;
+
+/*
+ * Created by stoneo on 1/11/2015.
+ */
+public class SlangFilesVerifier {
+
+    public static void main(String[] args) {
+        String repositoryPath = args[0];
+        ApplicationContext context = new AnnotationConfigApplicationContext(VerifierSpringConfiguration.class);
```
orius123: Would add a validation that path exists and handle valid. Plus maybe handling os path separators etc..

score-lang-content-verifier/src/main/java/org/openscore/lang/tools/verifier/SlangFilesVerifier.java
```diff
@@ -0,0 +1,35 @@
...
+        VerifierHelper verifierHelper = context.getBean(VerifierHelper.class);
+        try {
+            verifierHelper.verifyAllSlangFilesInDirAreValid(repositoryPath);
+            System.out.println("SUCCESS: All slang files under directory: \"" + repositoryPath + "\" are valid.");
+            System.exit(0);
+        } catch (Exception e) {
+            System.out.println(e.getMessage() + "\n\nFAILURE: Validation of slang files under directory: \"" + repositoryPath + "\" failed.");
+            System.exit(1);
```
orius123: Doesn't throwing an exception exit with error code anyway?

score-lang-content-verifier/src/main/java/org/openscore/lang/tools/verifier/SlangFilesVerifier.java
```diff
@@ -0,0 +1,35 @@
...
+ */
+public class SlangFilesVerifier {
+
+    public static void main(String[] args) {
+        String repositoryPath = args[0];
+        ApplicationContext context = new AnnotationConfigApplicationContext(VerifierSpringConfiguration.class);
+        VerifierHelper verifierHelper = context.getBean(VerifierHelper.class);
+        try {
+            verifierHelper.verifyAllSlangFilesInDirAreValid(repositoryPath);
+            System.out.println("SUCCESS: All slang files under directory: \"" + repositoryPath + "\" are valid.");
```
orius123: would move the messages to constants

score-lang-content-verifier/src/main/java/org/openscore/lang/tools/verifier/VerifierHelper.java
```diff
@@ -0,0 +1,134 @@
...
+
+    @Autowired
+    private SlangCompiler slangCompiler;
+
+    @Autowired
+    private ScoreCompiler scoreCompiler;
+
+    private final static Logger log = Logger.getLogger(SlangFilesVerifier.class);
+
+    private String[] SLANG_FILE_EXTENSIONS = {"sl", "sl.yaml", "sl.yml"};
```
orius123: what is `.sl.yml` and `.sl.yaml`? 
what about `.yaml` files?

score-lang-content-verifier/src/main/java/org/openscore/lang/tools/verifier/VerifierHelper.java
```diff
@@ -0,0 +1,134 @@
...
+     * Transform all Slang files in given directory to Slang models, and store them
+     * @param directoryPath given directory containing all Slang files
+     */
+    public void verifyAllSlangFilesInDirAreValid(String directoryPath){
+        transformSlangFilesInDirToModels(directoryPath);
+        compileAllSlangModels();
+    }
+
+    private void transformSlangFilesInDirToModels(String directoryPath) {
+        Validate.notNull(directoryPath, "Directory path path can not be null");
```
orius123: duplicate `path` in message. 
plus, don't we want to make it more CLI like errors? for example: `you must specify a path`

score-lang-content-verifier/src/main/java/org/openscore/lang/tools/verifier/VerifierHelper.java
```diff
@@ -0,0 +1,134 @@
...
+        Collection<File> slangFiles = FileUtils.listFiles(new File(directoryPath), SLANG_FILE_EXTENSIONS, true);
+        log.info("Start compiling all slang files under: " + directoryPath);
+        log.info(slangFiles.size() + " .sl files were found");
+        for(File slangFile: slangFiles){
+            Validate.isTrue(slangFile.isFile(), "filePath must lead to a file");
```
orius123: add the problemtaic file name, so one will be able to fix the issue

score-lang-content-verifier/src/main/java/org/openscore/lang/tools/verifier/VerifierHelper.java
```diff
@@ -0,0 +1,134 @@
...
+                sourceModel = slangCompiler.preCompile(fromFile(slangFile));
+            } catch (Exception e) {
+                log.error("Failed creating Slang ,models for directory: " + directoryPath + ". Exception is : " + e.getMessage());
+                throw e;
+            }
+            verifyStaticCode(slangFile, sourceModel);
+            slangModels.put(getUniqueName(sourceModel), sourceModel);
+        }
+        if(slangFiles.size() != slangModels.size()){
+            throw new RuntimeException("We found: " + slangFiles.size() + " .sl files in path: " + directoryPath + ". We managed to create slang models for only: " + slangModels.size());
```
orius123: 1. how can we get here?
  if something went wrong don't we throw an exception?
2. don't like the "we.." style message. again prefer CLI like messages.

score-lang-content-verifier/src/main/java/org/openscore/lang/tools/verifier/VerifierHelper.java
```diff
@@ -0,0 +1,134 @@
...
+    @Autowired
+    private ScoreCompiler scoreCompiler;
+
+    private final static Logger log = Logger.getLogger(SlangFilesVerifier.class);
+
+    private String[] SLANG_FILE_EXTENSIONS = {"sl", "sl.yaml", "sl.yml"};
+
+    private Map<String, Executable> slangModels = new HashMap<>();
+
+    private Map<String, CompilationArtifact> compiledArtifacts = new HashMap<>();
```
orius123: why do you hold it on the class state and don't return it from the method and pass to the next?

score-lang-content-verifier/src/main/java/org/openscore/lang/tools/verifier/VerifierHelper.java
```diff
@@ -0,0 +1,134 @@
...
+    private SlangCompiler slangCompiler;
+
+    @Autowired
+    private ScoreCompiler scoreCompiler;
+
+    private final static Logger log = Logger.getLogger(SlangFilesVerifier.class);
+
+    private String[] SLANG_FILE_EXTENSIONS = {"sl", "sl.yaml", "sl.yml"};
+
+    private Map<String, Executable> slangModels = new HashMap<>();
```
orius123: why do you hold it on the class state and don't return it from the method and pass to the next?

score-lang-content-verifier/src/main/java/org/openscore/lang/tools/verifier/VerifierHelper.java
```diff
@@ -0,0 +1,134 @@
...
+        for(Executable slangModel : models) {
+            try {
+                Set<Executable> dependenciesModels = getModelDependenciesRecursively(slangModel);
+                CompilationArtifact compiledSource = compiledArtifacts.get(slangModel.getName());
+                if (compiledSource == null) {
+                    compiledSource = scoreCompiler.compile(slangModel, dependenciesModels);
+                    log.info("Compiled: " + slangModel.getName() + " successfully");
+                    compiledArtifacts.put(slangModel.getName(), compiledSource);
+                } else {
+                    log.error("Failed to compile: " + slangModel.getName());
```
orius123: why? 
doesn't it just means that is was already compiled. (and not possible since it is a Set)

score-lang-content-verifier/src/main/java/org/openscore/lang/tools/verifier/VerifierHelper.java
```diff
@@ -0,0 +1,134 @@
...
+                }
+            } catch (Exception e) {
+                log.error("Failed compiling Slang source: " + slangModel.getName() + ". Exception is : " + e.getMessage());
+                throw e;
+            }
+        }
+        if(compiledArtifacts.size() != slangModels.size()){
+            throw new RuntimeException("Out of: " + slangModels.size() + " slang models, we managed to compile only: " + slangModels.size());
```
orius123: again how is it possible and message style.

orius123:
As I said i would change the file name to 
- Main/Bootstrap..
- SlangVerifier

score-lang-content-verifier/src/main/java/org/openscore/lang/tools/verifier/VerifierHelper.java
oritstone: @orius123 If for some reason the precompile() returned null, and did not throw an exception (as was the case for the system props), I think we want to get an error that not all files were compiled

score-lang-content-verifier/src/main/java/org/openscore/lang/tools/verifier/VerifierHelper.java
oritstone: Agree about the CLI issue. We should change it there and in the docs

score-lang-content-verifier/src/main/java/org/openscore/lang/tools/verifier/VerifierHelper.java
oritstone: Just making sure that the compiler did not return null for some compilation without throwing an exception

score-lang-content-verifier/src/main/java/org/openscore/lang/tools/verifier/SlangFilesVerifier.java
oritstone: @orius123 what do you mean?

orius123: nothing special 
just making it a `private final static String`

score-lang-content-verifier/src/main/java/org/openscore/lang/tools/verifier/VerifierHelper.java
orius123: cool

PR #107, opened 2015-02-15
Transformers shouldn't run for keys that does not exist

orius123:
Signed-off-by: Eliya Sadan <eliyasadan@gmail.com>


orius123:
@oritstone WDYT on this approach?
can't merge because tests fail..

oritstone:
I think it is OK. This way we don't get unused keys with null values. Also, it might prevent bug in some cases (null pointer exceptions in the transformers)

orius123:
ok cool.
so we need to fix the failing tests.

orius123:
@oritstone pls merge

PR #121, opened 2015-02-22
upgrade jython to beta 4

orius123:


orius123:
@avimora they upgraded jline version, can you check if it solves the dependecies issue? Can you check uptime also. If it's all fine please merge

PR #123, opened 2015-02-23
Change tasks to list

oritstone:


oritstone:
@orius123 , you can review & merge when you have the time

score-lang-compiler/src/main/java/org/openscore/lang/compiler/modeller/ExecutableBuilder.java
```diff
@@ -111,25 +111,34 @@ public Executable transformToExecutable(ParsedSlang parsedSlang, String execName
...
+                List<Map<String, Map<String, Object>>> workFlowRawData;
                 try{
-                    workFlowRawData = (LinkedHashMap) executableRawData.get(SlangTextualKeys.WORKFLOW_KEY);
+                    workFlowRawData = (List) executableRawData.get(SlangTextualKeys.WORKFLOW_KEY);
                 } catch (ClassCastException ex){
-                    throw new RuntimeException("Flow: '" + execName + "' syntax is illegal.\nBelow 'workflow' property there should be a map of tasks and not a list");
+                    throw new RuntimeException("Flow: '" + execName + "' syntax is illegal.\nBelow 'workflow' property there should be a list of tasks and not a map");
                 }
-                if (MapUtils.isEmpty(workFlowRawData)) {
+                if (workFlowRawData == null || workFlowRawData.size() == 0) {
```
orius123: Why don't you use collectionutils?

score-lang-compiler/src/main/java/org/openscore/lang/compiler/modeller/ExecutableBuilder.java
```diff
@@ -111,25 +111,34 @@ public Executable transformToExecutable(ParsedSlang parsedSlang, String execName
...
-                    onFailureData = (LinkedHashMap) workFlowRawData.remove(SlangTextualKeys.ON_FAILURE_KEY);
-                } catch (ClassCastException ex){
-                    throw new RuntimeException("Flow: '" + execName + "' syntax is illegal.\nBelow 'on_failure' property there should be a map of tasks and not a list");
-                }
-                if (MapUtils.isNotEmpty(onFailureData)) {
-                    onFailureWorkFlow = compileWorkFlow(onFailureData, imports, null, true);
+                List<Map<String, Map<String, Object>>> onFailureData;
+                Iterator<Map<String, Map<String, Object>>> tasksIterator = workFlowRawData.iterator();
+                while(tasksIterator.hasNext()){
+                    Map<String, Map<String, Object>> taskData = tasksIterator.next();
```
orius123: What happens if onfailure is not the first task in the workflow order, how can another step know it should navigate to it instead of the default failure step? I think it has to happen before compiling the workflow this is whhy I wrote this ugly implementation before.

score-lang-compiler/src/main/java/org/openscore/lang/compiler/modeller/ExecutableBuilder.java
```diff
@@ -111,25 +111,34 @@ public Executable transformToExecutable(ParsedSlang parsedSlang, String execName
                 if(!executableRawData.containsKey(SlangTextualKeys.WORKFLOW_KEY)){
                     throw new RuntimeException("Error compiling " + parsedSlang.getName() + ". Flow: " + execName + " has no workflow property");
                 }
-                LinkedHashMap<String, Map<String, Object>> workFlowRawData;
+                List<Map<String, Map<String, Object>>> workFlowRawData;
```
orius123: I think that the best representation is as a `list<tuple<string, map>` and since you don't have tuple in java you might use map.entry or guava's pair that extends it. It will save you all of the redundant iterator next to get the first key each time.

score-lang-compiler/src/main/java/org/openscore/lang/compiler/modeller/ExecutableBuilder.java
```diff
@@ -111,25 +111,34 @@ public Executable transformToExecutable(ParsedSlang parsedSlang, String execName
...
+                while(tasksIterator.hasNext()){
+                    Map<String, Map<String, Object>> taskData = tasksIterator.next();
+                    String taskName = taskData.keySet().iterator().next();
+                    if(taskName.equals(SlangTextualKeys.ON_FAILURE_KEY)){
+                        try{
+                            onFailureData = (List<Map<String, Map<String, Object>>>)taskData.values().iterator().next();
+                        } catch (ClassCastException ex){
+                            throw new RuntimeException("Flow: '" + execName + "' syntax is illegal.\nBelow 'on_failure' property there should be a list of tasks and not a map");
+                        }
+                        if (onFailureData != null && onFailureData.size() > 0) {
```
orius123: Again about collection utils

orius123:
Do you have an issue in slang content repo as well? Can you add a reference here so we won't forget it.

score-lang-compiler/src/main/java/org/openscore/lang/compiler/modeller/ExecutableBuilder.java
oritstone: The logic did not change, as far as I can tell...
Only the way to access the onFailure task and remove it changes, since it is now a list of maps

score-lang-compiler/src/main/java/org/openscore/lang/compiler/modeller/ExecutableBuilder.java
oritstone: Thanks, changed it before, but forgot to commit it :)

score-lang-compiler/src/main/java/org/openscore/lang/compiler/modeller/ExecutableBuilder.java
oritstone: I don't understand your comment...
The workFlow data is now a list of maps, not a map, and I need to look in the list for the map with the on_failure key, and also remove it from the list (which I do with the iterator). Do you have an example of how I can improve the code?

score-lang-compiler/src/main/java/org/openscore/lang/compiler/modeller/ExecutableBuilder.java
orius123: When you are right you are right.

oritstone:
@orius123, am I good to go or do you have other comments?
I already merged the loops changes from the master (a few conflicts...)

orius123:
If you fill comfortable with a map and not a entry/tuple, go ahead and merge.

PR #126, opened 2015-02-24
Validate all flow references results are wired to navigations

oritstone:


oritstone:
@orius123 please review & merge if OK

score-lang-compiler/src/main/java/org/openscore/lang/compiler/scorecompiler/ScoreCompilerImpl.java
```diff
@@ -82,6 +87,38 @@ public ExecutionPlan convert(Executable compiledExecutable) {
...
+            return;
+        }
+        Flow flow = (Flow)executable;
+        Deque<Task> tasks = flow.getWorkflow().getTasks();
+        for(Task task : tasks){
+            Map<String, String> taskNavigations = task.getNavigationStrings();
+            String refId = task.getRefId();
+            Executable reference = filteredDependencies.get(refId);
+            Validate.notNull(reference, "Cannot compile flow: \'" + executable.getName() + "\' since for task: \'" + task.getName()
+                                        + "\', the dependency: \'" + refId + "\' is missing.");
```
orius123: why do you print the refId if it's null??

Bonczidai:
I think the following issue is related to this PR:

```
 workflow:
    - task1 ... # has SUCCESS, FAILURE
    - task2 ... # has only SUCCESS
  results:
    - CUSTOM1: condition
    - CUSTOM2: condition
```

The compiler complains that the SUCCESS / FAILURE results of tasks are not wired.

How can I solve this issue? Am I missing something? Is it mandatory for a flow to contain SUCCESS / FAILURE results?

PR #135, opened 2015-03-03
Changed so that CLI supports specific extensions

### LINKED ISSUE #111: change CLI to support specific extenasions as does the Slang verifier
Support only:
- sl
- sl.yaml
- sl.yml

Do not support other extensions

tudorlesan:
Fixes #111

### DIFF
--- a/score-lang-cli/src/main/java/org/openscore/lang/cli/utils/CompilerHelperImpl.java
+++ b/score-lang-cli/src/main/java/org/openscore/lang/cli/utils/CompilerHelperImpl.java
@@ -14,6 +14,7 @@ import com.google.common.collect.Lists;
 
 import org.apache.commons.collections4.CollectionUtils;
 import org.apache.commons.io.FileUtils;
+import org.apache.commons.io.filefilter.SuffixFileFilter;
 import org.apache.commons.lang.Validate;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.log4j.Logger;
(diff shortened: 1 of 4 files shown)


tudorlesan:
fixes #111

score-lang-cli/src/main/java/org/openscore/lang/cli/utils/CompilerHelperImpl.java
```diff
@@ -62,6 +63,16 @@ public CompilationArtifact compile(String filePath, List<String> dependencies) t
         Set<SlangSource> depsSources = new HashSet<>();
         File file = new File(filePath);
         Validate.isTrue(file.isFile(), "File: " + file.getName() + " was not found");
+        boolean validFileExtension = false;
+
+        Collection<File> slangFiles = FileUtils.listFiles(new File(file.getParent()), SLANG_FILE_EXTENSIONS, false);
+        for (File slangFile : slangFiles) {
+            if (StringUtils.equals(slangFile.getAbsolutePath(), file.getAbsolutePath())){
+                validFileExtension = true;
```
orius123: why do you need this validation?

orius123:
tests?

score-lang-cli/src/main/java/org/openscore/lang/cli/utils/CompilerHelperImpl.java
tudorlesan: I was suggested to use FileUtils.listFiles() which gets a collection of files with a specified list of extensions from a directory. In our case we specify a file, so I execute listFiles() for the file's parent directory. Afterwards I have to check that my file is among the list of files returned by the method.

orius123: and if the user specifies --cp?

tudorlesan: --cp is checked afterwards.The logic is as follows. First it checks that the source file exists. Then it gets all the files from the parent directory with our specified extensions and looks for the source file within them. Afterwards it gets all the files from the classpath with those extensions

PR #152, opened 2015-03-09
execution path logic

### LINKED ISSUE #104: executionPath is not correct for ExecutableSteps.finish action
executionPath is not correct for ExecutableSteps.finishtExecutable action

for example 

for ExecutableSteps.startExecutable the execution path was 0/1/0
but for the same level  ExecutableSteps.finishtExecutable  the execution path was 0/1

-- i would prefer to use '.' instead of '/' as separator.

avimora:
fixes #104

### DIFF
--- a/score-lang-runtime/src/main/java/org/openscore/lang/runtime/steps/ExecutableSteps.java
+++ b/score-lang-runtime/src/main/java/org/openscore/lang/runtime/steps/ExecutableSteps.java
@@ -1,3 +1,11 @@
+/*
+ * (c) Copyright 2014 Hewlett-Packard Development Company, L.P.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Apache License v2.0 which accompany this distribution.
+ *
+ * The Apache License is available at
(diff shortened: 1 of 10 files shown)


avimora:
@orius123 please review & merge

score-lang-runtime/src/main/java/org/openscore/lang/runtime/steps/ExecutableSteps.java
```diff
@@ -57,8 +64,7 @@ public void startExecutable(@Param(EXECUTABLE_INPUTS_KEY) List<Input> executable
                                 @Param(NODE_NAME_KEY) String nodeName,
                                 @Param(NEXT_STEP_ID_KEY) Long nextStepId) {
         try {
-            runEnv.getExecutionPath()
-                  .down();
+//			runEnv.getExecutionPath().forward(); // Start with 1 for consistency
```
orius123: I think starting with `0` is fine.
pls remove the comment

orius123:
looks good.
a lot of check_weather fixes :smile: 
just remove the comment and merge.

PR #153, opened 2015-03-09
Validate name space not empty compiler

### LINKED ISSUE #146: precompile of flow without namespace returns wrong id
When calling slangCompiler.preCompile with a flow that doesn't have a namespace (the line is missing), and then calling getId() on the returned Executable, it returns "null.flowName".

meirwah:
to fix #146


score-lang-compiler/src/main/java/org/openscore/lang/compiler/scorecompiler/ScoreCompilerImpl.java
```diff
@@ -50,6 +50,7 @@
 
     @Override
     public CompilationArtifact compile(Executable executable, Set<Executable> path) {
+        Validate.notEmpty(executable.getNamespace(),"Operation/Flow must have a namespace");
```
meirwah: @oritstone / @orius123  is this apporach valid? or shoyld be a varifier thing only???

orius123: Looks great to me.

orius123: Just add the flow or operation name to the message

meirwah:
@oritstone will you merge this after short review?

meirwah:
@einatneu does this solves your issue ?

einatneu:
@meirwah - I need it to be in the precompile also

meirwah:
added @einatneu ... check now...

einatneu:
@meirwah - looks good, thanks!

PR #158, opened 2015-03-11
support any kind of map in loops

### LINKED ISSUE #143: For loops over maps from a file seem to not work
When trying to do a for loop on values from a system property file, or from an input file to the CLI, I couldn't make it work.
When parsing the resource file, we get a *LinkedHashMap* object. 
We can use in the slang files some iterable over it like: entrySet(), keys() or values(), but all are non Serilaizable.
So I couldn't find a way to make it work...

Bonczidai:
fixes #143 

### DIFF
--- a/score-lang-compiler/src/test/java/org/openscore/lang/compiler/modeller/transformers/ForTransformerTest.java
+++ b/score-lang-compiler/src/test/java/org/openscore/lang/compiler/modeller/transformers/ForTransformerTest.java
@@ -112,15 +112,6 @@ public class ForTransformerTest {
     }
 
     @Test
-    public void testValidMapStatementWithExpression() throws Exception {
-        ForLoopStatement statement = transformer.transform("k, v in dictionary.items()");
-        MapForLoopStatement mapForLoopStatement  = validateMapForLoopStatement(statement);
-        Assert.assertEquals("k", mapForLoopStatement.getKeyName());
-        Assert.assertEquals("v", mapForLoopStatement.getValueName());
-        Assert.assertEquals("dictionary.items()", statement.getCollectionExpression());
-    }
-
-    @Test
     public void testValidMapStatementAndTrim() throws Exception {
         ForLoopStatement statement = transformer.transform(" k, v   in  collection  ");
         MapForLoopStatement mapForLoopStatement  = validateMapForLoopStatement(statement);
--- a/score-lang-runtime/src/main/java/org/openscore/lang/runtime/bindings/LoopsBinding.java
+++ b/score-lang-runtime/src/main/java/org/openscore/lang/runtime/bindings/LoopsBinding.java
@@ -17,7 +17,6 @@ import org.openscore.lang.runtime.env.Context;
 import org.openscore.lang.runtime.env.ForLoopCondition;
 import org.openscore.lang.runtime.env.LoopCondition;
 import org.python.core.PyObject;
-import org.python.google.common.collect.Lists;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Component;
 import org.apache.commons.lang3.tuple.Pair;
@@ -27,6 +26,7 @@ import java.util.Arrays;
 import java.util.Map;
 import java.util.List;
 import java.util.Set;
+import java.util.ArrayList;
 import java.util.regex.Pattern;
 
 import static org.openscore.lang.runtime.env.LoopCondition.LOOP_CONDITION_KEY;
@@ -90,7 +90,7 @@ public class LoopsBinding {
 
         if (forLoopStatement instanceof MapForLoopStatement) {
             if (evalResult instanceof Map) {
-                List<Map.Entry<Serializable, Serializable>> entriesAsSerializable = Lists.newArrayList();
+                List<Map.Entry<Serializable, Serializable>> entriesAsSerializable = new ArrayList<>();
                 @SuppressWarnings("unchecked") Set<Map.Entry<Serializable, Serializable>> entrySet = ((Map) evalResult).entrySet();
                 for (Map.Entry<Serializable, Serializable> entry : entrySet) {
                     entriesAsSerializable.add(Pair.of(entry.getKey(), entry.getValue()));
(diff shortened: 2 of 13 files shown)


Bonczidai:
This approach simplifies the syntax from `k, v in map.items()` to `k, v in map` and works with any kind of object that implements the Map interface:

- when reading from file you get LinkedHashMap
- when using Python expression you get PyDictionary, both implementing Map

Internally we convert every Map to an Iterable object which is also Serializable -> an ArrayList of `org.apache.commons.lang3.tuple.Pair` elements.

@orius123 @oritstone let me know what you think..

orius123:
sounds good to me.
why do you need to make it some python/java concrete map?
the only thing I don't like here is that it defers from python syntax..

Bonczidai:
well, only Python 'maps' have `items()` method.. so if we want to support java maps we cannot stay with that syntax (maybe if we manually convert all java maps in the context to PyDictionary before looping task - I don't really like that approach)

oritstone:
:+1:

meirwah:
@Bonczidai please update branch...

Bonczidai:
updated..

meirwah:
Did you review this @orius123 @oritstone ?

oritstone:
@Bonczidai is it ready for review?

Bonczidai:
Yes

score-lang-compiler/src/test/java/org/openscore/lang/compiler/modeller/transformers/ForTransformerTest.java
```diff
@@ -112,15 +112,6 @@ public void testValidMapStatementWithoutSpaceAfterComma() throws Exception {
     }
 
     @Test
-    public void testValidMapStatementWithExpression() throws Exception {
```
orius123: why did you remove it?

score-lang-runtime/src/main/java/org/openscore/lang/runtime/bindings/LoopsBinding.java
```diff
@@ -16,15 +16,17 @@
 import org.openscore.lang.runtime.env.Context;
 import org.openscore.lang.runtime.env.ForLoopCondition;
 import org.openscore.lang.runtime.env.LoopCondition;
-import org.python.core.PyList;
 import org.python.core.PyObject;
-import org.python.core.PyTuple;
+import org.python.google.common.collect.Lists;
```
orius123: don't use jython shaded 3rd parties.
if you need guava add maven dep

score-lang-runtime/src/main/java/org/openscore/lang/runtime/bindings/LoopsBinding.java
```diff
@@ -86,10 +88,17 @@ private LoopCondition createForLoopCondition(ForLoopStatement forLoopStatement,
             throw new RuntimeException("Error evaluating for loop expression in task '" + nodeName + "', error is: \n" + t.getMessage(), t);
         }
 
-        if ((forLoopStatement instanceof MapForLoopStatement) && !(evalResult instanceof PyList)) {
-            throw new RuntimeException(
-                    "Invalid expression for iterating maps: " + collectionExpression
-                    + "\nValid format is: map.items()");
+        if (forLoopStatement instanceof MapForLoopStatement) {
+            if (evalResult instanceof Map) {
+                List<Map.Entry<Serializable, Serializable>> entriesAsSerializable = Lists.newArrayList();
```
orius123: String, Serializable?

score-lang-runtime/src/main/java/org/openscore/lang/runtime/bindings/LoopsBinding.java
```diff
@@ -86,10 +88,17 @@ private LoopCondition createForLoopCondition(ForLoopStatement forLoopStatement,
...
+                @SuppressWarnings("unchecked") Set<Map.Entry<Serializable, Serializable>> entrySet = ((Map) evalResult).entrySet();
+                for (Map.Entry<Serializable, Serializable> entry : entrySet) {
```
orius123: can't you construct a list from a set?

score-lang-compiler/src/test/java/org/openscore/lang/compiler/modeller/transformers/ForTransformerTest.java
Bonczidai: since we don't have `.items()` anymore, the tests simplifies to `k, v in dictionary` which is tested in  the testValidMapStatement() method

score-lang-runtime/src/main/java/org/openscore/lang/runtime/bindings/LoopsBinding.java
Bonczidai: Depends, do we want to admit non-String keys for maps in slang files: e.g. `k, v in { 1: "abc", 2: "efg" }` ? 

```
k=1
v="abc"
----------
k=2
v="egf"
```

score-lang-runtime/src/main/java/org/openscore/lang/runtime/bindings/LoopsBinding.java
Bonczidai: We need to make sure that `Pair` elements are created which are Serializable so we need to iterate the entrySet.. Maybe would be better to use lambda expression:

```
                @SuppressWarnings("unchecked") Set<Map.Entry<Serializable, Serializable>> entrySet = ((Map) evalResult).entrySet();

                List<Map.Entry<Serializable, Serializable>> entriesAsSerializable =
                        convert(entrySet, new Converter<Map.Entry<Serializable, Serializable>, Map.Entry<Serializable, Serializable>>() {
                    @Override
                    public Map.Entry<Serializable, Serializable> convert(Map.Entry<Serializable, Serializable> entry) {
                        return Pair.of(entry.getKey(), entry.getValue());
                    }
                });
```

WDYT?

Bonczidai: We have to stay with Pair elements, because `new ArrayList(entrySet)` returns an object that is not Serializable

Bonczidai:
updated..

meirwah:
can we merge this @orius123 ?

orius123:
@Bonczidai delete the branch if you don't need it anymore.

PR #159, opened 2015-03-12
declare before assign outputs

orius123:
Signed-off-by: Eliya Sadan <eliyasadan@gmail.com>

### DIFF
--- a/score-lang-runtime/src/main/java/org/openscore/lang/runtime/bindings/OutputsBinding.java
+++ b/score-lang-runtime/src/main/java/org/openscore/lang/runtime/bindings/OutputsBinding.java
@@ -50,7 +50,9 @@ public class OutputsBinding {
                     //put action outputs
                     scriptContext.putAll(actionReturnValues);
                     //declare the new output
-                    scriptContext.put(outputKey, actionReturnValues.get(outputKey));
+                    if (!actionReturnValues.containsKey(outputKey)) {
+                        scriptContext.put(outputKey, null);
+                    }
                     //put operation inputs as a map
                     if(MapUtils.isNotEmpty(inputs)) {
                         scriptContext.put(BIND_OUTPUT_FROM_INPUTS_KEY, (Serializable) inputs);
--- a/score-lang-runtime/src/main/java/org/openscore/lang/runtime/bindings/OutputsBinding.java
+++ b/score-lang-runtime/src/main/java/org/openscore/lang/runtime/bindings/OutputsBinding.java
@@ -49,6 +49,8 @@ public class OutputsBinding {
                     Map<String, Serializable> scriptContext = new HashMap<>();
                     //put action outputs
                     scriptContext.putAll(actionReturnValues);
+                    //declare the new output
+                    scriptContext.put(outputKey, actionReturnValues.get(outputKey));
                     //put operation inputs as a map
                     if(MapUtils.isNotEmpty(inputs)) {
                         scriptContext.put(BIND_OUTPUT_FROM_INPUTS_KEY, (Serializable) inputs);
--- a/score-lang-runtime/src/main/java/org/openscore/lang/runtime/bindings/OutputsBinding.java
+++ b/score-lang-runtime/src/main/java/org/openscore/lang/runtime/bindings/OutputsBinding.java
@@ -66,12 +66,10 @@ public class OutputsBinding {
                     }
                     //evaluate expression
 
-                    if (scriptResult != null) {
-                        try {
-                            outputs.put(outputKey, scriptResult);
-                        } catch (ClassCastException ex) {
-                            throw new RuntimeException("The output expression " + outputExpr + " does not return serializable value", ex);
-                        }
+                    try {
+                        outputs.put(outputKey, scriptResult);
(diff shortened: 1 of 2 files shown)


score-lang-runtime/src/main/java/org/openscore/lang/runtime/bindings/OutputsBinding.java
```diff
@@ -49,6 +49,8 @@
                     Map<String, Serializable> scriptContext = new HashMap<>();
                     //put action outputs
                     scriptContext.putAll(actionReturnValues);
+                    //declare the new output
+                    scriptContext.put(outputKey, actionReturnValues.get(outputKey));
```
oritstone: Better to add null to the scriptContext if the action return values do not contain it

score-lang-runtime/src/test/java/org/openscore/lang/runtime/bindings/OutputsBindingTest.java
```diff
@@ -78,18 +78,27 @@ public void testOperationOutputsNoExpressionMultipleOutputs() throws Exception {
...
-        Map<String, String> expectedOutputs = new HashMap<>();
+        Map<String, Serializable> expectedOutputs = new HashMap<>();
         expectedOutputs.put("host1", "valueHost1");
         expectedOutputs.put("host2", "valueHost2");
 
         Assert.assertEquals("Binding results are not as expected", expectedOutputs, result);
     }
 
+    @Test
+    public void testOperationOutputsNoExpressionAtAll() throws Exception {
```
oritstone: Does not throw exception

score-lang-runtime/src/test/java/org/openscore/lang/runtime/bindings/OutputsBindingTest.java
```diff
@@ -78,18 +78,27 @@ public void testOperationOutputsNoExpressionMultipleOutputs() throws Exception {
...
+        Map<String, Serializable> operationContext = prepareOperationContext();
+        Map<String, Serializable> actionReturnValues = new HashMap<>();
+        List<Output> outputs = Arrays.asList(createNoExpressionOutput("actionOutputKey1"));
+
+        outputsBinding.bindOutputs(operationContext, actionReturnValues, outputs);
+    }
```
oritstone: Add assert

oritstone:
Check if you can remove throws exception from tests

orius123:
@tudorlesan can you please test if this branch helps solving the content output issues?

tudorlesan:
@orius123 operations such as ssh_command.sl work since they just use simple outputs and so if they don't exist, they will get replaced with None. 

The problem is for more complex operations like create_openstack_server.sl which use outputs like:
- status_code: statusCode (so we have pythonic variables for operations that execute java actions)

For these kind of operations (there are alot of them) errors are still thrown. For example I'm trying to execute the openstack_health_check.sl flow but the openstack machine is not working(for some reason). The operation throws an exception and statusCode is not in the java action's outputs.

This statusCode is used both for the output(to be pythonic) and for the result (SUCCESS only if it is '202' so as expected the whole execution flow gets interrupted and a RuntimeException is thrown.

Also, flows like docker_images_maintenance which compute their flow outputs based on previously computed step outputs (which may not exist) still throw the same error and your solution of putting the flow output into the step and then assign "''" to it at the flow level doesn't seem to work.

So I'm not sure how to solve these, and it seems that  most of the flows/operations still need modifying. I really need to speak to you for further analysis on how to manage all these changes that need to be done.

orius123:
Regarding the openstack flow, I think that if the operation got to an exception that he can't recover from its valid for the flow to stop.

But in general I think that when accessing values that may not exit we cannot assume that they exit obviously..
A more safe approach would be `locals().get('optional_val')` or even `locals().get('optional_val', 'default').
@meirwah @oritstone WDYT?
Regarding the docker images flow, let's talk about it next week.

tudorlesan:
@orius123  I knew about the locals() solution, basically we could check each and every output is in locals() before using it, thought there will be lots of modifications. As for the openstack flow, imo, I think the flow should just end with failure, not throw RuntimeException if openstack server is down.

orius123:
Not all places just places that the var is optional.
About openstack, I guess java action should catch the exception and notify on some variable..
We don't have try catch at the moment.

tudorlesan:
The exception that gets to the CLI is not from the java action. The java action does thrown an error which gets printed to the execution.log and that's it. There would be no problem, but when that exception is thrown, the statusCode output doesn't get computed anymore and again we get the same error of outputs not existing. So we would have to figure out how to not assign the java action output to a python output if the first one was not computed.

orius123:
You can declare it first.
```yaml
- statusCode
- status_code: statusCode
```
With the change I made it will actually behaves like:
`locals().get('statusCode')

meirwah:
so we need to write each output twice?

oritstone:
I prefer to have our expressions with the *locals().get('optional_val')* syntax, and in the future consider making the syntax easier for these cases

orius123:
@oritstone recheck please

PR #161, opened 2015-03-12
Add system test support to content verifier

oritstone:


oritstone:
**Spec:**
- All slang content system tests will reside in slang-content repo under: *slang-content/test* dir
- All tests for a flow with the namespace a.b.c will reside in the directory: *slang-content/test/a/b/c*
- A test case will be a composition of an executable *.sl* file and a yaml file containing inputs & description
-

**Tasks**:
- [ ] Pack Slang runtime in the content verifier jar
- [ ]

oritstone:
**Tasks**:
- [ ] Pack Slang runtime in the content verifier jar
- [ ]

meirwah:
@oritstone what with negative tests (where we want to validate flow failed)
and we need some flag , for tests that need real Docker env (like pre-requiest)...

oritstone:
@meirwah good points.
I added to the spec above also the notion of test suites, in order to allow for running only the base test suite, or add, for example, running also Docker tests (if we have an env for it).
I also added to the spec support in system properties

meshipeer:
looks good - i dont sure i understand how you know which flow should run...

oritstone:
@meshipeer 
- For each test case there is a  test suite specified. If no test suite is specified, it means general.
When running the build tool (verifier) it will run all test cases in the general+ test suites that are given as a prameter.
- The name of the test cases file has a convention that it is identical to the name of the test flow to run + the *.inputs.yaml* extension. 

Does it answer your question?

meshipeer:
yes 10x

orius123:
@oritstone please see my latest changes and merge.

PR #164, opened 2015-03-16
clone run inputs

### LINKED ISSUE #160: schedule slang flow fails on inputs passed as unmodifiable collection
* also the run time exception is fired from Slang language and not from score as ScoreEvent.Error

orius123:
fixes #160


orius123:
@shajyhia does it solves you issue?

Issue #165, opened 2015-03-17
Outputs fix


tudorlesan:
Path logic was changed. The flow outputs were printed based on what the PATH key was. It was expecting "0/0" but now the actual value is 0. EXEC_START_PATH constant inside SyncTriggerEventListener class has to be changed.

oritstone:
@tudorlesan is the issue already fixed? Can we close it?

tudorlesan:
@oritstone yes it can be closed.

PR #166, opened 2015-03-17
Flow outputs fix

tudorlesan:
Changed EXEX_START_PATH constant to be consisted with the actual value from execution.log


orius123:
How did the tests passed?

tudorlesan:
Only the extractOutputs method is tested, it does not actually get values from the execution.log. A map is formed that represents the execution.log values and the actual values are the outputs extracted from that. Probably that has to be changed..

tudorlesan:
fixes #165

PR #170, opened 2015-03-19
Multi instance - compiler

Bonczidai:
For repository migration:
changes so far that can be merged into master.. Once this branch is merged, it can be deleted. I will continue to commit after we handled the migration.

adding @oritstone

### DIFF
--- a/score-lang-compiler/src/main/java/org/openscore/lang/compiler/modeller/transformers/AsyncLoopForTransformer.java
+++ b/score-lang-compiler/src/main/java/org/openscore/lang/compiler/modeller/transformers/AsyncLoopForTransformer.java
@@ -9,15 +9,12 @@
  *******************************************************************************/
 package org.openscore.lang.compiler.modeller.transformers;
 
-import org.apache.commons.lang.StringUtils;
 import org.openscore.lang.entities.AsyncLoopStatement;
 import org.openscore.lang.entities.ScoreLangConstants;
 import org.springframework.stereotype.Component;
 
 import java.util.Arrays;
 import java.util.List;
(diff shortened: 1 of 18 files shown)


score-lang-compiler/src/main/java/org/openscore/lang/compiler/modeller/transformers/AsyncLoopForTransformer.java
```diff
@@ -0,0 +1,82 @@
...
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+/**
+ * Date: 3/17/2015
+ *
+ * @author Bonczidai Levente
+ */
+@Component
+public class AsyncLoopForTransformer implements Transformer<String, AsyncLoopStatement> {
```
orius123: Can't we avoid some of the duplication with for transformer here?

Bonczidai: Hi,

to avoid code duplication, introduced the following abstractions:

- LoopStatement
    - AsyncLoopStatement
    - ForLoopStatement
        - ListForLoopStatement
        - MapForLoopStatement
        
and

- AbstractForTransformer
    - ForTransformer
    - AsyncLoopForTransformer

please check and let me know WDYT

orius123:
@Bonczidai only one problem..  it doesn't work. fails when I try to test it.. :wink:

Bonczidai:
Which test / use case is failing? They seem to work for me :bowtie:

Bonczidai:
moved to branch `async_loop` .. closing PR

PR #182, opened 2015-03-25
Update language name for execution plans

Bonczidai:


orius123:
@oritstone we need to update dev agenda

PR #190, opened 2015-03-29
Update Readme

meirwah:
Update readme example flow to be simple


meirwah:
I think that in the quickstart, example flow should be simpler, look at this change, WDYT?
@SamMarkowitz @meshipeer @orius123 ?

orius123:
Agree

meshipeer:
+1

PR #192, opened 2015-03-30
Throw RuntimeException to engine and let it handle it

### LINKED ISSUE #178: Runtime exception from Slang language does not reach score engine and no Failure score event is fired
Run time exception from Slang language does not reach score engine and no Failure score event is fired.

Score engine should receive or acknowledged when an exception is thrown from slang that cause stopping execution.

shajyhia:
fixes #178


meshipeer:
fixes #178

PR #198, opened 2015-03-31
adding compatibility with windows os

raskin-aleksandr:
Flow changed to be compatible with windows os

### DIFF
--- a/build/build_content/build_cli_flow.sl
+++ b/build/build_content/build_cli_flow.sl
@@ -28,7 +28,7 @@ flow:
         default: 'target_dir + "/cloudslang-cli/cslang"'
         overridable: false
     - cloudslang_content_repo:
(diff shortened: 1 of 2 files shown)


build/build_content/build_cli_flow.sl
```diff
@@ -26,7 +28,7 @@ flow:
         default: 'target_dir + "/cloudslang-cli/cslang"'
         overridable: false
     - cloudslang_content_repo:
-        default: "'https://github.com/CloudSlang/cloud-slang-content.git'"
```
orius123: why did you move to SSH?

meirwah:
It seems that now on win version "cslang-cli.zip" is saved under the "cli" folder (instead of target)
@raskin-aleksandr @orius123 is that what you meant?

orius123:
What. Ever works. Doesn't really matter.

Issue #154, opened 2015-03-09
unstable loops tests


meirwah:
The tests : 
  LoopFlowsTest.testFlowWithMapLoopsWithDefaultBreak:152 expected:<2> but was:<3>
  LoopFlowsTest.testFlowWithMapLoops:122 null
  LoopFlowsTest.testFlowWithMapLoopsWithBreak:181 expected:<3> but was:<4>

Fail locally at my machine (noth by running from maven AND from IDE)...

meirwah:
@Bonczidai / @orius123 are you familier with this ?

orius123:
I have sporadic failures, not related to maps loops.

meirwah:
I get this tests fail each time... not sporadic ...

meirwah:
Please look at this issue @Bonczidai ...

meirwah:
Failed tests:
  LoopFlowsTest.testFlowWithMapLoopsWithDefaultBreak:164 expected:<2> but was:<3>
  LoopFlowsTest.testFlowWithMapLoops:118->verifyPersonMap:206 null
  LoopFlowsTest.testFlowWithHashMap:136->verifyPersonMap:206 null
  LoopFlowsTest.testFlowWithMapLoopsWithBreak:193 expected:<3> but was:<4>

PR #202, opened 2015-04-01
add travis to test  Against Multiple JDKs(7+8)

meirwah:
Signed-off-by: Meir Wahnon <meir.wahnon@hp.com>


meirwah:
This is pending on #154

PR #208, opened 2015-04-02
added coveralls and jacoco plugins

orius123:
Using coveralls for monitoring our code coverage.
we can set rules in coveralls for failing for desegregation.
I would recommend we should merge it after we fix the problems it points to.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: java
 
 after_success:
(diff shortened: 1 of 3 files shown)


meshipeer:
license Apache 2

orius123:
License?

meshipeer:
i just checked that its OK for us to use it...

Issue #193, opened 2015-03-30
cloudSlang started to run and then got stuck


gseffy:
I tried to run a simple flow from docker image. The flow started to run and then got stuck after:
Database name is: H2 databaseMajorVersion is: 1

The log file:
2015-03-30 11:05:41:435 0    [main] INFO  org.springframework.beans.factory.xml.XmlBeanDefinitionReader  - Loading XML bean definitions from URL [jar:file:/cslang/cslang/lib/cloudslang-cli-0.7.3.jar!/META-INF/spring/spring-shell-plugin.xml]
2015-03-30 11:05:41:811 376  [main] INFO  org.springframework.beans.factory.xml.XmlBeanDefinitionReader  - Loading XML bean definitions from URL [jar:file:/cslang/cslang/lib/cloudslang-cli-0.7.3.jar!/META-INF/spring/slangCliContext.xml]
2015-03-30 11:05:41:943 508  [main] INFO  io.cloudslang.schema.ScoreNamespaceHandler  - Registering Score namespace handler
2015-03-30 11:05:42:026 591  [main] INFO  org.springframework.beans.factory.xml.XmlBeanDefinitionReader  - Loading XML bean definitions from class path resource [META-INF/spring/score/context/scoreRepositoryContext.xml]
2015-03-30 11:07:49:336 127901 [main] WARN  org.springframework.beans.factory.xml.XmlBeanDefinitionReader  - Ignored XML validation warning
org.xml.sax.SAXParseException; systemId: http://www.springframework.org/schema/tx/spring-tx.xsd; lineNumber: 12; columnNumber: 150; schema_reference.4: Failed to read schema document 'http://www.springframework.org/schema/tool/spring-tool-4.1.xsd', because 1) could not find the document; 2) the document could not be read; 3) the root element of the document is not <xsd:schema>.
        at com.sun.org.apache.xerces.internal.util.ErrorHandlerWrapper.createSAXParseException(ErrorHandlerWrapper.java:198)
        at com.sun.org.apache.xerces.internal.util.ErrorHandlerWrapper.warning(ErrorHandlerWrapper.java:99)
        at com.sun.org.apache.xerces.internal.impl.XMLErrorReporter.reportError(XMLErrorReporter.java:433)
        at com.sun.org.apache.xerces.internal.impl.xs.traversers.XSDHandler.reportSchemaErr(XSDHandler.java:4162)
[...]

PR #231, opened 2015-04-06
update spring version to 4.1.4 (to match score version)

meirwah:
Signed-off-by: wahnonm <meir.wahnon@hp.com>

### DIFF
--- a/cloudslang-tests/src/main/java/io/cloudslang/lang/systemtests/TriggerFlows.java
+++ b/cloudslang-tests/src/main/java/io/cloudslang/lang/systemtests/TriggerFlows.java
@@ -66,7 +66,19 @@ public class TriggerFlows {
         RunDataAggregatorListener listener = new RunDataAggregatorListener();
         slang.subscribeOnEvents(listener, STEP_EVENTS);
 
+        try {
+            Thread.sleep(2000L);      /* TODO : remove this! only to test unstable navigation tests*/
+        } catch (InterruptedException e) {
+            e.printStackTrace();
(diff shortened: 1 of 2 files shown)


meirwah:
fixes #193 and #167 ...
On the long term, we need to see how make sure spring versions at score and cslang are aligned, and how we add test for this (maybe part of system tests?)

meirwah:
@meshipeer  or @orius123  any reason why i should not update spring version to be like in score?

meirwah:
seems that test : testDefaultSuccessNavigation in NavigationTest fails, @Bonczidai  can you take a look at that?

meirwah:
seems that on Linux machines this happens:
java.lang.NullPointerException: null
	at io.cloudslang.lang.systemtests.flows.NavigationTest.testDefaultSuccessNavigation(NavigationTest.java:137)

caused by:

2015-04-06 14:28:08 ERROR ExecutionServiceImpl:100 - Error during execution: 
java.lang.NullPointerException
	at io.cloudslang.score.events.EventBusImpl.dispatch(EventBusImpl.java:38)

meshipeer:
i don't see a reason the spring versions shouldn't be the same. lets make sure the tests are stable and than merge it

meirwah:
I don't understand why the push build passes, while the pr fails.. very weird.. it should be the same code in both...

orius123:
maybe sporadic surprises..
we should use aligned versions of spring projects (like spring IO offers) in both score and cloud-slang

PR #237, opened 2015-04-09
validate required inputs

### LINKED ISSUE #224: Required input can be passed as empty
1. Have an operation with an input but no value
2. Run it and do not provide --i
Result: you are required to run it as error appears: 
Command failed java.lang.RuntimeException: Slang Error : Error running: 'fileListParser': Input with name: 'fileList' is Required, but value is empty
3. Provide input with --i but with no actual value

Expected result: just input name and no value should not make the flow continue
Actual result: flow finishes with empty required input!

Bonczidai:
Fixes #224 

### DIFF
--- a/cloudslang-runtime/src/main/java/io/cloudslang/lang/runtime/bindings/InputsBinding.java
+++ b/cloudslang-runtime/src/main/java/io/cloudslang/lang/runtime/bindings/InputsBinding.java
@@ -1,23 +1,25 @@
 package io.cloudslang.lang.runtime.bindings;
 
 /*******************************************************************************
-* (c) Copyright 2014 Hewlett-Packard Development Company, L.P.
-* All rights reserved. This program and the accompanying materials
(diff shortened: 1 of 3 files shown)


Bonczidai:
the validation cases for required inputs:

- null value
- empty string
- empty collection

meirwah:
:+1:

meirwah:
I think we have an issue here...
Let's look at this op for example : 
https://github.com/CloudSlang/cloud-slang-content/blob/master/content/io/cloudslang/coreos/get_machine_public_ip.sl

Where we have an input of:
- privateKeyFile:
        default: "''"

This is by defualt requiered=true, means it will fail for sure...

We have similar inputs all over our content... so we need to be carful about this PR (maybe wait to have System tests merged? )...

Maybe keep it for now? and fix later ?

meirwah:
Or fix this in all the content...

PR #239, opened 2015-04-12
Update Readme

meirwah:
update cli version


orius123:
I think we should direct to our site / download and update the link over there

meirwah:
Also...

orius123:
The link is in index.js in site repo

PR #240, opened 2015-04-12
Update Readme

meirwah:
remove score mention...

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,8 @@
 CloudSlang
 ==============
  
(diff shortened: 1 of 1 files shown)


SamMarkowitz:
The "Score" was left in on purpose because the repo is still called **score** and in the docs we sometimes use "Score" as a shorthand for CloudSlang Orchestration Engine. 

@meirwah, if you still think it should be removed, I'll merge.

meirwah:
I think it should still be removed... @oritstone @meshipeer WDYT?

PR #256, opened 2015-04-15
Add unit tests to the system tests framework

oritstone:


oritstone:
@orius123 You can review & merge what I did so far
Still lots more tests to be done, but it helps with the coverage a bit... :)

PR #277, opened 2015-04-22
Add unit tests to the SlangBuildMain class

oritstone:


oritstone:
@orius123 have a look.
I don't like the tests. Do you have an idea how to test it in another way?
Also, running the tests takes a lot of time...

PR #315, opened 2015-06-07
Allow builder to run w\o default test suite

### LINKED ISSUE #296: System tests fw - support running only specific test suite (w\o)base

oritstone:
fixes #296
Allow running the builder without running the default test suite - by specifying testSuite: !default

### DIFF
--- a/cloudslang-content-verifier/src/main/java/io/cloudslang/lang/tools/build/SlangBuildMain.java
+++ b/cloudslang-content-verifier/src/main/java/io/cloudslang/lang/tools/build/SlangBuildMain.java
@@ -38,8 +38,10 @@ public class SlangBuildMain {
 
     private static final String CONTENT_DIR =  File.separator + "content";
     private static final String TEST_DIR = File.separator + "test";
+    public static final String DEFAULT_TESTS = "default";
 
     private final static Logger log = Logger.getLogger(SlangBuildMain.class);
+    private static String NOT_TS = "!";
 
     public static void main(String[] args) {
         ApplicationArgs appArgs = new ApplicationArgs();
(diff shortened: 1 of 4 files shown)


cloudslang-content-verifier/src/main/java/io/cloudslang/lang/tools/build/SlangBuildMain.java
```diff
@@ -91,6 +93,20 @@ public static void main(String[] args) {
         }
     }
 
+    private static List<String> parseTestSuites(ApplicationArgs appArgs) {
+        boolean runDefaultTests = true;
+        List<String> testSuites = ListUtils.defaultIfNull(appArgs.getTestSuites(), new ArrayList<String>());
+        for(String testSuite : testSuites){
```
orius123: why not just appending any suite that does't start with `!` ?

cloudslang-content-verifier/src/main/java/io/cloudslang/lang/tools/build/SlangBuildMain.java
```diff
@@ -91,6 +93,20 @@ public static void main(String[] args) {
...
+            if(testSuite.equalsIgnoreCase(NOT_TS + DEFAULT_TESTS)){
+                runDefaultTests = false;
+            }
+        }
+        if(runDefaultTests && !testSuites.contains(DEFAULT_TESTS)){
```
orius123: what happens when I pass `ts=!default,default` ?

oritstone: What do you think should happen? Currently it will be treated as running the default test suite

cloudslang-content-verifier/src/main/java/io/cloudslang/lang/tools/build/SlangBuildMain.java
oritstone: @orius123 good idea. Anyway, regarding the default test suite, I still need to have a special treatment, since we want to add it even if it is not specifically written

PR #319, opened 2015-06-15
added help and handle unkown params

orius123:


orius123:
@oritstone can we merge?

PR #326, opened 2015-06-18
try to exclude for coverage

orius123:


orius123:
@oritstone WDYT?

oritstone:
@orius123 why do you want to exclude SlangBuildMain? I think we do need tests for it...

PR #337, opened 2015-06-24
When no default value is specified, do not put the name of the input as it's expression 

### LINKED ISSUE #336: preCompile API returns input object which contains input name as default value
When the input has required / overridable / encrypted  field (the value doesn’t matter), and has no default value, the expression field of Input object contains the input name during the precompile API.

oritstone:
Fixes #336 
When no default value is specified, do not put the name of the input as it's expression (for the case where the input has additional properties)

### DIFF
--- a/cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/modeller/transformers/AbstractInputsTransformer.java
+++ b/cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/modeller/transformers/AbstractInputsTransformer.java
@@ -68,7 +68,7 @@ public abstract class AbstractInputsTransformer {
                 (boolean) props.get(OVERRIDABLE_KEY);
         boolean defaultSpecified = props.containsKey(DEFAULT_KEY);
         String inputName = entry.getKey();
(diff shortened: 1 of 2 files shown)


cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/modeller/transformers/AbstractInputsTransformer.java
```diff
@@ -68,8 +68,7 @@ private static Input createPropInput(Map.Entry<String, Map<String, Serializable>
                 (boolean) props.get(OVERRIDABLE_KEY);
         boolean defaultSpecified = props.containsKey(DEFAULT_KEY);
         String inputName = entry.getKey();
-        String expression = defaultSpecified ? props.get(DEFAULT_KEY)
-                                                    .toString() : inputName;
+        String expression = props.containsKey(DEFAULT_KEY) ? (String)props.get(DEFAULT_KEY).toString() : null;
```
orius123: Y R U `!` re-using the `defaultSpecified` var?

PR #347, opened 2015-06-29
handle errors within branches

### LINKED ISSUE #330: Join branches throws NPE when aggregating wrong output
when running flow:
https://gist.github.com/orius123/a5c2e9facf20e6d0de00

NPE is thrown from AsyncLoopSteps
my best guess is that something in this chain in null
[Relevant line](https://github.com/CloudSlang/cloud-slang/blob/7c2ff9c81ad706e065bf394f890b31c55ad6c038/cloudslang-runtime/src/main/java/io/cloudslang/lang/runtime/steps/AsyncLoopSteps.java#L224)

10x @genadi-hp for catching this issue

orius123:
fix #330

### DIFF
--- a/cloudslang-runtime/src/main/java/io/cloudslang/lang/runtime/steps/AsyncLoopSteps.java
+++ b/cloudslang-runtime/src/main/java/io/cloudslang/lang/runtime/steps/AsyncLoopSteps.java
@@ -132,7 +132,7 @@ public class AsyncLoopSteps extends AbstractSteps {
             Map<String, Serializable> contextBeforeSplit = flowContext.getImmutableViewOfVariables();
             List<String> branchesResult = Lists.newArrayList();
 
-            collectBranchesData(runEnv, executionRuntimeServices, nodeName, branchesContext, branchesResult);
+            collectBranchesData(executionRuntimeServices, nodeName, branchesContext, branchesResult);
 
             Map<String, Serializable> publishValues =
                     bindAggregateOutputs(
@@ -223,7 +223,6 @@ public class AsyncLoopSteps extends AbstractSteps {
     }
 
     private void collectBranchesData(
-            RunEnvironment runEnv,
             ExecutionRuntimeServices executionRuntimeServices,
             String nodeName,
             List<Map<String, Serializable>> branchesContext,
--- a/cloudslang-runtime/src/main/java/io/cloudslang/lang/runtime/steps/AsyncLoopSteps.java
+++ b/cloudslang-runtime/src/main/java/io/cloudslang/lang/runtime/steps/AsyncLoopSteps.java
@@ -46,6 +46,7 @@ import static io.cloudslang.score.api.execution.ExecutionParametersConsts.EXECUT
 public class AsyncLoopSteps extends AbstractSteps {
 
     public static final String BRANCH_EXCEPTION_PREFIX = "Error running branch";
+    private static final String BRANCH_ID = "BRANCH_ID";
 
     @Autowired
     private AsyncLoopBinding asyncLoopBinding;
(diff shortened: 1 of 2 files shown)


Bonczidai:
added a minor commit (just extracted a variable) 
Problem was: we get an exception inside the branch from the input binding since `fromInputs` is not supported for inputs (just outputs, results)

with @orius123's fix the error message is more descriptive:
```
Command failed java.lang.RuntimeException: Slang Error : Error running: 'print_values':
Error running branch:
Error running: print_values: Error binding input: 'text',
    Error is: Error in running script expression or variable reference, for expression: 'fromInputs['input_' + str(value)]',
    Script exception is: javax.script.ScriptException: NameError: name 'fromInputs' is not defined in <script> at line number 1
```

btw the flow also contains other errors: according to flow logic range should be `(0, 3)`, flow result has expression

is there anything else we want to check regarding this issue?

oritstone:
Add unit tests

Bonczidai:
that's a good point ..

Bonczidai:
added unit test

adding system-test for it involves changes in the *systemtests* classes since now the execution stops after the first Exception (from the branch) and does not get to the Exception from the *join* step - so I don't think we should do that now (but if you think otherwise - let me know)

also, the error log message in the *join* step prints the branch ID (*executionRuntimeServices.getBranchId()*) which will always be *null* since it is called in the parent branch - @orius123 , is there any specific reason for leaving this in the message?

```
  logger.error("There was an error running branch: " +
                        executionRuntimeServices.getBranchId() + " Error is: " + branchException);
```

orius123:
But it is inside a loop that iterates over the branches.

cloudslang-runtime/src/main/java/io/cloudslang/lang/runtime/steps/AsyncLoopSteps.java
```diff
@@ -217,6 +220,15 @@ private void collectBranchesData(
 
         List<EndBranchDataContainer> branches = executionRuntimeServices.getFinishedChildBranchesData();
         for (EndBranchDataContainer branch : branches) {
+
+            //first we check that no exception was thrown during the execution of the branch
+            String branchException = branch.getException();
+            if (StringUtils.isNotEmpty(branchException)) {
+                logger.error("There was an error running branch: " +
+                        executionRuntimeServices.getBranchId() + " Error is: " + branchException);
+                throw new RuntimeException(BRANCH_EXCEPTION_PREFIX + ": \n" + branchException);
```
oritstone: Do we want the behavior to be that in case one of the branches fail we throw an exception and stop execution?

orius123:
yes

Bonczidai:
@orius123 just to be sure, now the error log looks like this:

```
2015-07-13 12:11:29:397 11582 [1_WorkerExecutionThread-4_101600001] ERROR io.cloudslang.lang.runtime.steps.AsyncLoopSteps  - There was an error running branch: null Error is: Error running: print_values: Error binding input: 'text', 
	Error is: Error in running script expression or variable reference, for expression: 'fromInputs['input_' + str(value)]',
	Script exception is: javax.script.ScriptException: NameError: name 'fromInputs' is not defined in <script> at line number 1
```

branch ID null, is this what we want?

orius123:
you are right. we should take the branch id from the EndBranchDataContainer executionRuntimeServices and not from the parent.
But than we need null checks all over..

oritstone:
@Bonczidai are you fixing it?

Bonczidai:
on it

Bonczidai:
fixed.. now the ID looks like this: `BRANCH_ID=2029f616-1a45-4578-9bca-2a0a6825ad7a:2`

please review

PR #354, opened 2015-07-06
Double qoute fix

### LINKED ISSUE #226: I bumped into the ‘double quota’ issue. I couldn’t overcome it on my own

orius123:
fixes #226


orius123:
**This PR is not ready for merge ATM. we need to fix all of the test and content in advance.**

Bonczidai:
updated most of the tests - still, there are some cases that I want to check (they are ignored right know)

orius123:
@Bonczidai  I think we should add on the code a comment explaining how this fix works. I prefer you do so we won't have conflicts. 
@meirwah @SamMarkowitz we should explain the functionality impact very good in the docs since this is our only deviation from the YAML spec.

Bonczidai:
Currently we cannot write an expression like: 'string' + var

`#1`

`- first_string: "prefix_" + first` throws:

```
in 'string', line 22, column 15:
                - first_string: "prefix_" + first
                  ^
expected <block end>, but found Scalar
 in 'string', line 22, column 39:
     ...       - first_string: "prefix_" + first
                                         ^
```

`#2` 
`- first_string: '"prefix_" + first'` interprets it as a string not an expression

.. yet, this form works:

`#3`

```
            - second_string:
                default: >
                  'prefix_' + second_string
```

orius123:
This is a pretty hacky solution. If we are sure we want to go with it and not with some marker for expressions or something like it, than we need a lot of testing here.

Bonczidai:
also adding @SamMarkowitz @tudorlesan @mihaitusa @gituser173

Bonczidai:
ping @meirwah -- need to sync about this

meirwah:
@lrevnic @avigail-oron WDYT on this issue?

orius123:
I think that from functionality POV, we should make it work exactly like a bar assignment will work in Python/Java. Meaning that `input: "123"` is a string, and `input: "123" + "456"` is an expression. 
Don't know how to do it technically, but that is what I would expect as a user.

oritstone:
I don't think that in terms of usability\functionality it is acceptable to demand the author to write a one-line expression with a > sign and multi-lines.
Changing the Yaml parser this way is hacky, and seems to me to be dangerous.
Maybe we should go with another solution, as @orius123 stated above of marking the expressions in a specific way. 
e.g. `input: ${"prefix_" + first}` or `input: ->"prefix_" + first`

oritstone:
After some further investigation, we understood that the problem with parsing the expression above:
```yaml
- first_string: 'prefix_' + first
```
is with the need for escaping the character `+`
We can achieve that for example by demanding the author to precede expressions with special characters by escaping char. For example:
 ```yaml
- first_string: $'prefix_' + first
```
or:
 ```yaml
- first_string: ->'prefix_' + first
```
The issue is, if it is acceptable to expect it from the authors of CloudSlang.

To support that we would still need hacking the Yaml Parser for adding additional set of quotes to strings constants in order to distinguish between them and expressions.

orius123:
So we will stay with the hack and add new marker for expressions. 
Technically it sound horrible for me. Functional wise, I think this is a good approach. Users will have to get used to / migrate for using marker for expressions but I believe it adds clarity to the language. Every thing else is a string.

orius123:
Actually do we need the hack if we demand a marker?

orius123:
@oritstone @Bonczidai @meirwah

meirwah:
I do not like at all the tech implication of this...
I'm afraid that we both won't be Yaml compatible and usability wise it might not be worth it...

oritstone:
In the comment I wrote above, I did not mean a marker for expressions, we can still use:
```yaml
- first_string: firstString
```
which is technically also an expression without any escaping.
We will need to use `$` or `->` only when we use special characters in the expression such as `+`
This way we will get simpler language for the common,simple cases, and add escaping only in the cases the YAML itself demand the escaping.

And then as far as I understand we will still need the hack to distinguish between string constants and expressions.
We will still use the python interpreter for everything.

orius123:
If it is just about special characters, can't we tell the user to escape them in normal ways, like `\+` or wrapping with quotes?
And than we give a consistent story.

Bonczidai:
#### summary:

```yaml
- input: flow_input
- input: > ...
- input: |
```

not affected by the hack..

-----------------------------------------------

fixed by the hack:

```yaml
- input: 'bla' # STR
- input: "bla" # STR
```

side effect of the hack:

```yaml
- input: "'a' + b" # STR
```

-----------------------------------------------

problem seem to be when you **start** the *value* with a quote (because it ends the value when the quote ends and Yaml parser complaints) - e.g.:

```yaml
- input: a - b # ok
- input: a - 'b' # ok
- input: 'a' - b # NOT ok
```

**but** we can do something like this:

```yaml
- input: \'a' - b # ok
- input: \'a\' - b # ok
```

Bonczidai:
#### summary (complementary to the previous one):

current format:
```yaml
    - int_input: 6
    - bool_input: false
    - str_input: "'abc'"
    - from_var: var_name
    - expr1: return_result if return_code == '-1' else ''
    - expr2: standard_out[:-1]
    - expr3: "'a' + var_name"
```

format with the hack:
```yaml
    - int_input: 6
    - bool_input: false
    - str_input: 'abc'
    - from_var: var_name
    - expr1: return_result if return_code == '-1' else ''
    - expr2: standard_out[:-1]
    - expr3: >
        'a' + var_name
```

without the hack, marker for expressions:

```yaml
    - int_input: 6
    - bool_input: false
    - str_input1: abc
    - str_input2: 'abc'
    - from_var: -> var_name
    - expr1: -> return_result if return_code == '-1' else ''
    - expr2: -> standard_out[:-1]
    - expr3: -> 'a' + var_name
```

Bonczidai:
if we go with the third one, we should also decide whether to use the marker (`->`) for output expressions / result conditions

Bonczidai:
@CloudSlang/slang-functional-reviewers 

What are your opinions on this issue?

avigail-oron:
I prefer marking expressions rather than marking strings with double quotes. 
The reason is that in the expression option, authors will understand what is expected (even if it's awkward and takes more time) while in the double quotes option I would not understand that i'm expected to do that:  " ' somthing' + myVar" 
and not " 'something' " + myVar

oritstone:
I discussed the issue also with @lirantal.
He states that marking the expressions specifically, for example using `->` is also confusing, and also something you need to get used to, so he does not see the big advantage over the current situation.
In addition, it might be more common than we think when using YAML - look at http://rundeck.org/docs/manual/jobs.html#quoting-arguments-to-steps

oritstone:
We can try another approach - doing some pre-processing over the initial YAML file, before parsing it using the YAML parser, in order to add quotes to the entire value where needed.

For example, Change:
```yaml
    - str_input: 'abc'
```
to:
```yaml
    - str_input: "'abc'"
```
And:
```yaml
    - expr3: 'a' + var_name
```
to
```yaml
    - expr3: "'a' + var_name"
```
And only then send it to the YAML parser

SamMarkowitz:
I think it will still need some explanation because it differs from the official YAML spec in at least two ways:

+ strings don't need any quotes
```
  - str_input: abc
```
+ this is not valid YAML
```
  - expr3: 'a' + var_name
```

meirwah:
I איןמל אישא 'ק טםו דועעקדא @oritstone is complex, and has any end-cases...

orius123:
Still think that the right approach is to have the user mark the type ( string or expression).
If -> is confusing we can go with other signs. 
Maybe wrap with () or {{}}

lirantal:
@orius123 can you give a practical example of using () and {} rather than the double quotes?

orius123:
```yaml 
input1: some string
input2: (input1)
input3:{{input1 + ", another string"}}
```
Results in:
input1 = some string
input2 = some string
input3 = some string, another string

Didn't check it from YAML perspective but maybe I missed some escaping chars. 

And of course I don't mean we should use both options, but we should choose either `()` or `{{}}` or anything else of course...

lirantal:
@oritstone @Bonczidai 

So about this... Here's an ansible example (based on YAML too):
https://github.com/geerlingguy/ansible-for-devops/blob/master/nodejs/provisioning/playbook.yml#L48

Looks like they have " and ' as well.
They also have a {{ }} notation for variables which is indeed easier to understand and is widely used in AngularJS, some templating engines like swig, and other frameworks I guess.

orius123:
Just mentioning that our use of the `{{}}` might be a little different than ansible's, we don't only mention references in there we also evaluate expressions. @lirantal In angular you can right JS expression in the `{{}}` ?

oritstone:
@orius123 , you are right about Ansible, it might be a bit confusing since Ansible only support references using this syntax, for example `{{input1}}`.
In Angular JS, however, this syntax supports expressions such as: `{{input1 + input2}}`

I think there is an advantage for using:
```yaml
input3: {{input1 + input2}}
``` 
instead of:
```yaml
input3: -> input1 + input2
```
since it mark the end of the expression as well, which might be more comprehensive when the expression takes more than one line

oritstone:
@mbettan @benc-uk WDYT? Does the double quotes syntax bug you? Do you think the notion of marking expressions with `{{}}` is better?

lirantal:
IMHO I think `{{}}` makes more sense than the `->` thing, possibly partly because it's a familiar pattern from other frameworks/languages, and also @oritstone's comments about understanding where it begins/ends which I didn't think about but it makes a lot of sense when someone says it out loud.

benc-uk:
I'm trying to follow the whole thread, it's a little difficult in places :confused:

I much prefer `{{}}` to the `->`  syntax, the arrow is too much like a pointer dereference or property access (e.g. in PHP) the braces makes a lot more sense.

Would supporting this mean we no longer need to double quote when defining a static string literal? If so, I very much support this! It can be confusing for new learners of CloudSlang, and when using input files which are *plain* YAML you do not need to double quote (in fact it causes errors) so you end up with a confusing mix

mbettan:
Same from my side, I agree with with @benc-uk

benc-uk:
Thinking about this. Will this mean we will have two ways of including expressions within a flow?
The `>` syntax for multiline and `{{}}` for inline? What about using the `>` syntax for both. E.g
```yaml
input: >
  "foo" + bar
```
and
```yaml
input: > "foo" + bar
```
Just an idea! :smile:

oritstone:
@orius123 @lirantal @avigail-oron @tomershadi @benc-uk @mbettan
Actually I just tested it and:
```yaml
input2: {{"hi" + input1}}
```
Is not valid YAML
And neither is:
```yaml
input2: > "hi" + input1
```
I guess that because the two characters have specific meaning in YAML syntax...

So if we choose to use expressions, we need to have some other marking.
For example:
```yaml
input2: -> "hi" + input1
```
or:
```yaml
input2: ${"hi" + input1}
```
WDYT?

benc-uk:
Whoops, didn't realise that `>` was part of the YAML specification (as is `|`), I thought it was an extension unique to the CloudSlang parser. 
Looks like `{` character [is already used by YAML](http://www.yaml.org/refcard.html) to denote a "in-line keyed branch" 

I also notice that the `{{}}` syntax is [used by Ansible, to denote a variable](http://docs.ansible.com/ansible/YAMLSyntax.html) but only when it's inside a quoted string

So maybe `${"hi" + input1}` is a good alternative, it will be familiar to many people including OO users

orius123:
(Some expression) like bash maybe

lirantal:
I think that's still confusing.
What other options can we have?

oritstone:
@lirantal do you find `${expr}`confusing because it usually refers to substitution of a parameter and not expression?

How about using `[[expr]]` \ `#[expr]` \ `$[expr]` \ `#{expr}`?
Other suggestions?

benc-uk:
Square brackets are used by YAML already, and so is hash `#` for comments
I think dollar is pretty much the only option left :confused:

I don't think it's confusing BTW

oritstone:
Thanks everyone for the productive debate :)
We will go with marking expressions using `${expr}`.
For eample:
```yaml
- input1: ${input3}
...
- output1: ${input7 + "hi"}
```
And make constant values syntax simpler:
```yaml
- input1: "hi"
```

benc-uk:
Good stuff, this will make things a lot clearer to new users I think
:clap: :+1:

orius123:
:(

orius123:
I like the string simplicity. 
but `${}` is not a convention anywhere, and even in OO it means something else.
I think it will be even more confusing for OO users.

oritstone:
@orius123 do you have another idea?
We decided to go with marking expressions differently, in order to achieve this simplicity and better authoring experience. 
So now it is only a matter of syntax.
If you have a better idea regarding the syntax, we will appreciate it a lot :)

orius123:
I think that `()` is good but it looks like I'm a minority here.. As a developer I don't find `$` intuitive. It's not common in the programming langs / frameworks I worked with.

orius123:
My guess OO took this syntax from bash. So lets go bash over here with ()

lirantal:
@oritstone @meshipeer I know this is a closed thread but I was thinking - 
If we anyway need to re-write all the content now to accommodate the double quote fix, did you ever considered moving from YAML to JSON notation? JSON is definitely the de-facto today for today's developers ecosystem.

orius123:
I don't think json is de-facto in the automation and devops stuff. YAML is pretty strong in this area.
anyway supporting both (as long as they are valid) doesn't seem like a big effort IMHO.

mbettan:
I agree - Customers are talking only about YAML nowadays

lirantal:
You're both right, common DevOps tool do use YAML, although my comment was with regards to developers nowadays in the AngularJS, NodeJS, and general hype in the JavaScript ecosystem.

benc-uk:
Yeah plenty of JS hype, but people are using JSON as intended - to
serialise objects, they aren't writing JSON documents by hand. JSON is a
nightmare to "code" in, you're neck deep in [] and {} before you even
start. I've been there!
YAML is much more human friendly for actually creating and editing

lirantal:
Well that is arguable because node packages are maintained with package.json, as well as bower.json, and while you're right and these usually don't grow that big and are fairly simple to handle, SublimeText IDE uses JSON to manage the IDE configuration.

That said, I still agree with you that YAML is the best case for DevOps ecosystem.

PR #358, opened 2015-07-07
added possible solution to compiler error message

### LINKED ISSUE #220: compiler error - Error message unclear when no alias is provided for an operation/flow import.
@SSgtMendel during bug hunt

### LINKED ISSUE #222: compiler error - The compilation error not clear is due to missing space after the import operation
bughunt:bughunt.errors.compilation

tudorlesan:
Fixes #220 
Fixes #222

### DIFF
--- a/cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/parser/YamlParser.java
+++ b/cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/parser/YamlParser.java
@@ -46,7 +46,7 @@ public class YamlParser {
                 errorMessage += "Probably did not provide (key: value) pair or missing space after colon(:)";
             }
             if (e instanceof ConstructorException && errorMessage.startsWith("Cannot create property")){
-                errorMessage += "Are you sure you are not missing a space after colon(:) for one of the imports?";
+                errorMessage += "Are you sure you are not missing a space after colon(:) for one of the imports and that all imports have an alias?";
             }
 
             throw new RuntimeException("There was a problem parsing the YAML source: " + source.getName() + ".\n" + errorMessage, e);
--- a/cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/parser/YamlParser.java
+++ b/cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/parser/YamlParser.java
@@ -26,6 +26,7 @@ import org.yaml.snakeyaml.scanner.ScannerException;
 @Component
 public class YamlParser {
 
+    public static final String KEY_VALUE_PAIR_MISSING_MSG = "Probably did not provide (key: value) pair or missing space after colon(:). Also check that everything is indented properly";
     @Autowired
     private Yaml yaml;
 
@@ -43,7 +44,7 @@ public class YamlParser {
         } catch (Throwable e) {
             String errorMessage = e.getMessage();
             if (e instanceof ScannerException && (errorMessage.startsWith("mapping values") || errorMessage.startsWith("while scanning a simple key"))){
-                errorMessage += "Probably did not provide (key: value) pair or missing space after colon(:)";
+                errorMessage += KEY_VALUE_PAIR_MISSING_MSG;
             }
             else if (e instanceof ConstructorException && errorMessage.startsWith("Cannot create property")){
                 if (errorMessage.contains("Unable to find property")){
@@ -52,7 +53,7 @@ public class YamlParser {
                     errorMessage += "Property \'" + undefinedProperty + "\' is not supported by CloudSlang. Check that \'" + undefinedProperty + "\' is indented properly.";
                 }
                 else if (errorMessage.contains("No single argument constructor found for interface java.util.Map")){
-                    errorMessage += "Probably did not provide (key: value) pair or missing space after colon(:)";
+                    errorMessage += KEY_VALUE_PAIR_MISSING_MSG;
                 }
             }
             throw new RuntimeException("There was a problem parsing the YAML source: " + source.getName() + ".\n" + errorMessage, e);
--- a/cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/parser/YamlParser.java
+++ b/cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/parser/YamlParser.java
@@ -26,7 +26,15 @@ import org.yaml.snakeyaml.scanner.ScannerException;
 @Component
 public class YamlParser {
 
-    public static final String KEY_VALUE_PAIR_MISSING_MSG = "Probably did not provide (key: value) pair or missing space after colon(:). Also check that everything is indented properly";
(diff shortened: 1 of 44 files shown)


tudorlesan:
Since #220 and #222 are not more detailed, I think this solves both issues.

cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/parser/YamlParser.java
```diff
@@ -39,7 +41,15 @@ public ParsedSlang parse(SlangSource source) {
             parsedSlang.setName(source.getName());
             return parsedSlang;
         } catch (Throwable e) {
-            throw new RuntimeException("There was a problem parsing the YAML source: " + source.getName() + ".\n" + e.getMessage(), e);
+            String errorMessage = e.getMessage();
+            if (e instanceof ScannerException && errorMessage.startsWith("mapping values")){
+                errorMessage += "Probably did not provide (key: value) pair or missing space after colon(:)";
+            }
+            if (e instanceof ConstructorException && errorMessage.startsWith("Cannot create property")){
```
meirwah: why not "else if"?

cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/parser/YamlParser.java
```diff
@@ -39,7 +41,15 @@ public ParsedSlang parse(SlangSource source) {
             parsedSlang.setName(source.getName());
             return parsedSlang;
         } catch (Throwable e) {
-            throw new RuntimeException("There was a problem parsing the YAML source: " + source.getName() + ".\n" + e.getMessage(), e);
+            String errorMessage = e.getMessage();
+            if (e instanceof ScannerException && errorMessage.startsWith("mapping values")){
+                errorMessage += "Probably did not provide (key: value) pair or missing space after colon(:)";
```
meirwah: errorMessage +=  ?
I think the idea is that you write the specific error msg ("Probably did not provid..") before the general one no?

cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/parser/YamlParser.java
tudorlesan: I figured "e" cannot be two error types at once. But I can add "else if" just to be more clear

cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/parser/YamlParser.java
tudorlesan: Well I was thinking the actual problem was presented first, then a possible solution could be printed. Like in C# there is the error message: The type or namespace name 'a_namespace' could not be found (are you missing a using directive or an assembly reference?).

cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/parser/YamlParser.java
```diff
@@ -39,7 +41,15 @@ public ParsedSlang parse(SlangSource source) {
...
+            }
+            else if (e instanceof ConstructorException && errorMessage.startsWith("Cannot create property")){
+                errorMessage += "Are you sure you are not missing a space after colon(:) for one of the imports and that all imports have an alias?";
```
oritstone: The exception message starts with: *Cannot create property=imports* and I think you should also the imports text when checking, since you only address errors in imports...

oritstone:
Missing unit tests for these test cases. Add tests to: CompilerErrorsTest

cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/parser/YamlParser.java
oritstone: I'm OK with this order

oritstone:
This fix actually deals only with errors in the first alias of the imports.
If the problem is with the second alias, such as:
```yaml
imports:
  consul: io.cloudslang.consul
  consul2:aa
``` 
Then your fix does not catch it.

Not sure we have a way to catch it, but please check

cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/parser/YamlParser.java
tudorlesan: I did not fully understand this. Could you elaborate please?

tudorlesan:
The same exception (ScannerException) is thrown in that case only it contains a different message. Just had to include that message in the condition and it catches it

cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/parser/YamlParser.java
oritstone: Your error message is specific to imports. If this is the case, you should include the property name (imports) in the check of the exception message ("Cannot create property=imports").
In your tests you tested a more general use case, where the error could come from any property for which a space is missing. So maybe it is better to leave the if statement the same, but make the error message more general. Not only for the imports use case

cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/parser/YamlParser.java
```diff
@@ -39,7 +41,21 @@ public ParsedSlang parse(SlangSource source) {
...
+                if (errorMessage.contains("Unable to find property")){
+                    String truncatedErrorMessage = errorMessage.substring(errorMessage.indexOf("Unable"), errorMessage.indexOf("on class"));
+                    String undefinedProperty = truncatedErrorMessage.substring(truncatedErrorMessage.indexOf("\'")+1, truncatedErrorMessage.lastIndexOf("\'"));
+                    errorMessage += "Property \'" + undefinedProperty + "\' is not supported by CloudSlang. Check that \'" + undefinedProperty + "\' is indented properly.";
+                }
+                else if (errorMessage.contains("No single argument constructor found for interface java.util.Map")){
+                    errorMessage += "Probably did not provide (key: value) pair or missing space after colon(:)";
+                }
+            }
+            throw new RuntimeException("There was a problem parsing the YAML source: " + source.getName() + ".\n" + errorMessage, e);
```
oritstone: @tudorlesan In the message thrown in the runtime exception, the exception error message appears twice - once from the e.getMessage() and once from the errorMessage.
You should only throw the exception message once.

orius123:
@tudorlesan can you please extract constant values and add comments to the code. 
Since there is a lot of searching through strings here it's really hard to follow from someone that is out of context (like me..)

tudorlesan:
@orius123 I extracted the constant values and added some comments on the code, tests, and test flows to associate tests, testflows and errors. Basically there are 4 types of errors covered in this PR. The code checks each error message to establish what specific message to add. 3 of them contain the same message being added, while for the other, a parsing is made to be able to print what property is undefined. Each testflow now also has a comment saying what the problem is. Tell me if you still have questions.

orius123:
@tudorlesan it looks great.
just one small OCD comment.. can you please make the tests use the same constants please.. :grin:

PR #359, opened 2015-07-08
Added  message to indicate how to include a system property file

### LINKED ISSUE #221: If a system property is present, but there is no system property file, error message doesn’t mention that it may be a missing system property file
@SamMarkowitz during bug hunt

tudorlesan:
Fixes #221 

### DIFF
--- a/cloudslang-cli/src/main/java/io/cloudslang/lang/cli/services/ScoreServicesImpl.java
+++ b/cloudslang-cli/src/main/java/io/cloudslang/lang/cli/services/ScoreServicesImpl.java
@@ -87,6 +87,9 @@ public class ScoreServicesImpl implements ScoreServices{
 
         String errorMessageFlowExecution = scoreEventListener.getErrorMessage();
         if (StringUtils.isNotEmpty(errorMessageFlowExecution)) {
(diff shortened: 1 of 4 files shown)


tudorlesan:
fixex #221

oritstone:
@tudorlesan I do not think this fix is correct. In this case, if there are any system properties defined in the content, and one of the required inputs in missing (even if it does not contain the sys prop attribute), you will receive your error message.
I think your fix should be in the InputBinding class - if a required input value is missing, and it has a sys property attribute defined but no value in the sys properties

tudorlesan:
Command failed java.lang.RuntimeException: Slang Error : Error running: 'ping_hosts'.
     Input with name: 'ip_list' is Required, but value is empty
This value can also be supplied using a system property     << Added from InputsBinding if a missing input has "system_property:"

A system property file can be included using --spf <path_to_file>   << Added from CLI if the error message contains the previous message.

tudorlesan:
@oritstone tell me if that's ok

oritstone:
@tudorlesan the fix is OK
You need to add unit tests for these cases

PR #364, opened 2015-07-12
Fire a new event for starting input binding from executable & from task

### LINKED ISSUE #349: missing event INPUT_START_ACTION for start input binding
today we have only INPUT_END_ACTION event thrown from Clang and it is not enough for cases of an exception happen at binding level

oritstone:
Fixes #349


oritstone:
@Bonczidai please review & merge

Bonczidai:
:+1: 

travis jdk8 build was failing sporadically - works after restart (opened issue here so we will be aware of it if it happens again: #365 )

PR #369, opened 2015-07-14
support step type (operation ,subflow,other)

### LINKED ISSUE #341: Missing data about the task type in CloudSlang events
missing data about task type - if it runs a flow or operation

shajyhia:
Fixes #341


orius123:
Ok. after talking to @liat-bensal we want this information to be stored on the Executable and not on the tasks. (that way we can also have a type for branch)
I suggest we put it on the stepType key.
@shajyhia does it help you with your use-case?

shajyhia:
i am using stepType.  i need it on Task and not executable(call me for more questions) .

meirwah:
can we close this ? @shajyhia @oritstone ?

liat-bensal:
no

orius123:
Just emphasizing again that putting it in the task is a bad approach, this is not a data of the task.
Putting it in the executable is different implementation that the current one.

orius123:
can we close this one?

PR #370, opened 2015-07-14
Added possible solution to null value input

### LINKED ISSUE #199: Uninformative error message in CLI when missing spaces before 'required' keyword
When there are missing spaces before input's 'required' keyword, for example:
flow:
  inputs:
    - input:
      required: true

 and trying to run the flow in the CLI, you get an uninformative error message:
Command failed java.lang.RuntimeException: java.lang.RuntimeException: Error transforming source: flow_with_required_input.sl to a Slang model. null

tudorlesan:
fixes #199


tudorlesan:
fixes #199

orius123:
@oritstone & @tudorlesan 
I think this fix is great, but I don't think it solves #199.
for example:
```yaml
inputs:
  input1:
  required: false
    default: "aaa"
```
And many other cases will still not be clear to the user..

tudorlesan:
@orius123  As I see, this case is "mapping values are not allowed here" which is treated in #358. An error message proposing some possible mistakes along with a suggestion to check everything is indented properly is added.

PR #372, opened 2015-07-15
Added error message for case where inputs file or sp file is empty

### LINKED ISSUE #344: When inputs file is empty a NPE is thrown to sout
should print some informative error

tudorlesan:
fixes #344 

### DIFF
--- a/cloudslang-cli/src/main/java/io/cloudslang/lang/cli/utils/CompilerHelperImpl.java
+++ b/cloudslang-cli/src/main/java/io/cloudslang/lang/cli/utils/CompilerHelperImpl.java
@@ -129,12 +129,10 @@ public class CompilerHelperImpl implements CompilerHelper{
                 String inputsFileContent = FileUtils.readFileToString(new File(inputFile));
                 Boolean emptyContent = true;
                 if (StringUtils.isNotEmpty(inputsFileContent)) {
-                    emptyContent = false;
                     @SuppressWarnings("unchecked") Map<String, ? extends Serializable> inputFileYamlContent =
                             (Map<String, ? extends Serializable>) yaml.load(inputsFileContent);
-                    if (MapUtils.isEmpty(inputFileYamlContent)) {
-                        emptyContent = true;
-                    } else {
+                    if (MapUtils.isNotEmpty(inputFileYamlContent)) {
+                        emptyContent = false;
                         result.putAll(inputFileYamlContent);
                     }
                 }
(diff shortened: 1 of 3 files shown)


tudorlesan:
fixes #344

Bonczidai:
<< correct me if I am wrong >> but it seems to me that this code section doesn't do its job

`throw new RuntimeException("Specified inputs file or system property file is empty or does not contain valid YAML content.");` - this will never be thrown: if the inputs file is empty the execution passes silently because of this line: `if (StringUtils.isNotEmpty(inputsFileContent)) {`

if the inputs file syntax is incorrect (e.g. contains a list not a map) an uninformative ClassCastException will be thrown

`if (inputFileYamlContent != null) {` becomes redundant because of `(StringUtils.isNotEmpty(inputsFileContent)) {` - anyway I think MapUtils.isNotEmpy() would be better here

I would restructure this section a little bit: check first the loading of the files (one specific error), then check if its empty and if it contains a valid map of inputs (custom error here not ClassCastException)

and of course - unit tests :boom: 

@tudorlesan  WDYT?

orius123:
Added file name to error

orius123:
The error will be thrown when you have only commented out lines in your yaml  file

tudorlesan:
@Bonczidai I wouldn't have pushed this if I didn't check it before and make sure the error is thrown. You are right that if the file is truly empty execution will pass silently, but this NPE was thrown if it had commented code as @orius123 said. As for the MapUtils.isNotEmpty() this checks both if the map is null or just empty. If the return was an empty map, then the NPE would not be thrown, that's why I was thinking to check only if it is null.

So basically the logic is like this. First check if the file is truly empty - then execution passed silently. If it's not empty, try to load yaml content from it. If it succeeds loading any yaml content , be it with syntax errors or not, it will either return a non-null map or throw a different kind of error. Finally, if yaml.load() returns a null map, this is the case that this issue specifically treats. It throws a different error specifying that the file might be empty or contain invalid content (comments, or maybe other characters that make .load() return null) instead of just a NPE.

tudorlesan:
@Bonczidai tell me if you still disagree with some of the points I made and and I can change the code. Though restructuring the logic seems like interfering with issues that are not related with this particular branch

orius123:
after thinking it through, I think that we should treat empty file and commented out file / empty map.. (file with no actual content) the same @tudorlesan WDYT?

tudorlesan:
@orius123 yes, I can do that. Though commented out file produces a null map. I'm actually not sure what yaml content would produce an empty map..

orius123:
Me 2

Issue #311, opened 2015-06-03
Add --addcp param to CLI


meirwah:
--cp will replace default cp location depdency.
--addcp will be to add to existing content defult location additional cp locations...

It should support multiple folder locations.

PR #404, opened 2015-08-04
Added --addcp param

tudorlesan:


tudorlesan:
fixes #311

tudorlesan:
I think the tests have to be changed since adding any other file in the io.cloudslang.lang.cli test package makes other tests fail and not specifying a classpath gets the default value as the whole test folder, so it doesn't make sense to check if the additional classpath was added since it was clearly added in the default value.

cloudslang-cli/src/main/java/io/cloudslang/lang/cli/SlangCLI.java
```diff
@@ -63,14 +63,15 @@
 
     @CliCommand(value = "run", help = "triggers a CloudSlang flow")
     public String run(
-            @CliOption(key = {"", "f", "file"}, mandatory = true, help = "Path to filename. e.g. cslang run --f C:\\CloudSlang\\flow.yaml") final File file,
-            @CliOption(key = {"cp", "classpath"}, mandatory = false, help = "Classpath , a directory comma separated list to flow dependencies, by default it will take flow file dir") final List<String> classPath,
+            @CliOption(key = {"", "f", "file"}, mandatory = true, help = "path to filename. e.g. cslang run --f C:\\CloudSlang\\flow.yaml") final File file,
```
meirwah: why this has changed?

cloudslang-cli/src/main/java/io/cloudslang/lang/cli/SlangCLI.java
```diff
@@ -63,14 +63,15 @@
...
+            @CliOption(key = {"cp", "classpath"}, mandatory = false, help = "classpath , a directory comma separated list to flow dependencies, by default it will take flow file dir") final List<String> classPath,
```
meirwah: why this has changed?

cloudslang-cli/src/main/java/io/cloudslang/lang/cli/SlangCLI.java
```diff
@@ -103,10 +104,11 @@ public String setEnvVar(
 
     @CliCommand(value = "inputs", help = "Get flow inputs")
     public List<String> getFlowInputs(
-            @CliOption(key = {"", "f", "file"}, mandatory = true, help = "Path to filename. e.g. cslang inputs --f C:\\CloudSlang\\flow.yaml") final File file,
-            @CliOption(key = {"cp", "classpath"}, mandatory = false, help = "Classpath , a directory comma separated list to flow dependencies, by default it will take flow file dir") final List<String> classPath)
+            @CliOption(key = {"", "f", "file"}, mandatory = true, help = "path to filename. e.g. cslang inputs --f C:\\CloudSlang\\flow.yaml") final File file,
+            @CliOption(key = {"cp", "classpath"}, mandatory = false, help = "classpath , a directory comma separated list to flow dependencies, by default it will take flow file dir") final List<String> classPath,
```
meirwah: if this is "a directory comma separated list to flow dependencies"
so what is addcp?

cloudslang-cli/src/main/java/io/cloudslang/lang/cli/SlangCLI.java
tudorlesan: I just lowercase-ed the first word to be consistent with the other options. Should I change it back?

cloudslang-cli/src/main/java/io/cloudslang/lang/cli/SlangCLI.java
tudorlesan: same as above

cloudslang-cli/src/main/java/io/cloudslang/lang/cli/SlangCLI.java
tudorlesan: same, but the --addcp is used, it will add the directories to the default classpath instead of replacing it

meirwah:
testing is not enough....
where is the end/corner use-cases?
where you have both CP flag and ADDCP flag turned on?
what if you have same folder in both?
what if you have default folder in ADDCP ? will it work?

tudorlesan:
Bad news:

2015-08-05 13:00:55 ERROR SlangImpl:60 - Failed compilation for source : flow_with_missing_navigation_from_op_result.sl ,Exception is : Cannot compile flow: 'basic_flow' since for task: 'Task1', the result 'CUSTOM' of its dependency: 'user.ops.print_custom_result_op' has no matching navigation
Tests run: 7, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 1.238 sec <<< FAILURE! - in io.cloudslang.lang.systemtests.SimpleFlowTest
testFlowWithGlobalSession(io.cloudslang.lang.systemtests.SimpleFlowTest)  Time elapsed: 0.028 sec  <<< ERROR!

Compilation error yet rebuild succeeded.

Issue #407, opened 2015-08-09
Update file flow question


nirgur:
Hi
I'm trying to create a flow that updates a file
This flow has 2 steps,
1. read the file's content
2. Write an updated content to the file

the problem is that on the write task, the flow doesn't use the updated value, it's using the value from the flow's input so instead of updating the file's content it's just overriding the content with my input
Maybe i'm missing something
this is the flow:

```yaml

namespace: update_file.flow
imports:
  files: io.cloudslang.base.files

flow:
  name: files_and_folders
  
  inputs:
   - file_path
   - text
  
  workflow:
    - ReadFile:
        do:
         files.read_from_file:
               - file_path
        publish:
          - read_text: read_text

    - UpdateFile:
        do:
               files.write_to_file:
               - file_path
               - text: "read_text + '\n' + text" 
```

meirwah:
why you need "text" as flow input , if you are not using it???

nirgur:
why not? see last line

meirwah:
anyway you need to use the: 
overridable: false

since by default , the inputs are "assignedFrom=true"  , it will take the input from the existing context...

meirwah:
see example at : http://www.cloudslang.io/#/docs#inputs

nirgur:
i'm not sure I understand
which input I need to set as overridable: false??
My problem is that the "text" input of the write_to_file operation is using the flow's "text" input instead of the concatenated string

meirwah:
input "text" in task "UpdateFile", that way it will become "assignedFrom=false" , and use the right context for the value

orius123:
I think it's #288 all over again...
make the task `text` input:
```yaml
- text: 
    default: "read_text + '\n' + text"
    overridable: false
```

or change the flow inputs to something like
`suffix` / `extra_text` / `user_addition` ...

oritstone:
@nirgur does it answer your question?
The real issue is as mentioned in #288. We currently treat task arguments exactly like inputs, and not as arguments to run a subflow or operation.
We plan to address that issue on the next version

oritstone:
After fixing #288 this issue does not happen anymore

PR #414, opened 2015-08-11
Fix: same input name for flow and task

### LINKED ISSUE #288: Task arguments act like inputs
for example they perform the required logic and in cast that a required input is not available fail the flow instead of passing the value as-is

### LINKED ISSUE #410: Error message when flow & operation has the same input name
Run this flow and see that you're getting an error message on the input text
change the flow's input name from text to something else and see that everything is working

```yaml
namespace: update_file.flow
imports:
  files: io.cloudslang.base.files
  print: io.cloudslang.base.print
  utils: io.cloudslang.base.utils

flow:
  name: update_file
  
  inputs:
   - file_path
   - text
  
  workflow:
    - ReadFile:
        do:
         files.read_from_file:
         - file_path
        publish:
          - read_text
         
    - UpdateFile:
        do:
         files.write_to_file:
         - file_path
         - text:
            default: "read_text + '\\n' + text"
            overridable: false
            
    - Sleep:
        do:
         utils.sleep:
         - seconds: 0
```

Bonczidai:
fixes #410 and also _fixing #288 , right @orius123 ?

### DIFF
--- a/cloudslang-runtime/src/main/java/io/cloudslang/lang/runtime/bindings/InputsBinding.java
+++ b/cloudslang-runtime/src/main/java/io/cloudslang/lang/runtime/bindings/InputsBinding.java
@@ -75,17 +75,20 @@ public class InputsBinding {
         targetContext.put(inputName, value);
     }
 
-    private Serializable resolveValue(Input input, Map<String, ? extends Serializable> context,
-                                      Map<String, ? extends Serializable> targetContext,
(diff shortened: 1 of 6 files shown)


orius123:
it doesn't fully fix #288
it does fix #407 though

orius123:
For #288 we want to make task argument a new type. not to act like input at all.

PR #418, opened 2015-08-12
OOM issue: pass only relevant runtime environment data to branches

Bonczidai:

### DIFF
--- a/cloudslang-runtime/src/main/java/io/cloudslang/lang/runtime/env/RunEnvironment.java
+++ b/cloudslang-runtime/src/main/java/io/cloudslang/lang/runtime/env/RunEnvironment.java
@@ -11,6 +11,7 @@
 package io.cloudslang.lang.runtime.env;
 
 import com.hp.oo.sdk.content.plugin.SerializableSessionObject;
+import org.apache.commons.lang.SerializationUtils;
 
 import java.io.Serializable;
 import java.util.HashMap;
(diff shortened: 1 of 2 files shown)


cloudslang-runtime/src/main/java/io/cloudslang/lang/runtime/env/RunEnvironment.java
```diff
@@ -107,4 +108,36 @@ public ExecutionPath getExecutionPath() {
...
+        Map<String, SerializableSessionObject> serializableDataMapCopy =
+                (Map<String, SerializableSessionObject>) SerializationUtils.clone((Serializable) serializableDataMap);
+        runEnvironmentWithoutStacks.serializableDataMap = serializableDataMapCopy;
+
+        runEnvironmentWithoutStacks.returnValues =
+                (ReturnValues) SerializationUtils.clone(returnValues);
+
+        runEnvironmentWithoutStacks.nextStepPosition = nextStepPosition;
+
+        return runEnvironmentWithoutStacks;
```
meirwah: this is a bit tricky, since if someone , add/changes RunEnvironment  Object he must update this method (cloneWithoutStacks) otherwise weird stuff will happen... what about reverse approach? 
clone normally and just reset (after that) the ContextStack field.
It's more work (less efficient)  but code wise, might be cleaner....

@orius123  WDYT?

Bonczidai: in that case we should add setter methods in RunEnv for the ContextStack and ParentFlowStack  fields / add `reset` methods to the ..Stack classes

`.. @orius123 WDYT?`

orius123: We can add a `resetStacks` method and go @meirwah way.
@Bonczidai WDYT?

Bonczidai: :+1:

meirwah:
:+1:

orius123:
@meirwah do we want it also in 0.8 branch?

meirwah:
yes.. but first verify the OOM issue does not reproduce (in OO) , and then merge it...

orius123:
We need to merge it so OO will be able to test it..

meirwah:
ok ,please merge it with branch (0.8)

PR #426, opened 2015-08-14
Async loop: refactor event names

### LINKED ISSUE #355: Async Loop events names are wrong
Currently i am using EVENT_ASYNC_LOOP_EXPRESSION_END and EVENT_ASYNC_LOOP_OUTPUT_END to support Async Loop but these two event names does not represent the action it self (JOIN,SPLIT) and could be missing some data about the number and names  of the branches and so.

i will recommend using the following names "EVENT_SPLIT_BRANCHES" and "EVENT_JOIN_BRANCHES"

Bonczidai:
fixes #355 


meirwah:
@shajyhia will this affect OO?

orius123:
OO doesn't need it. It's for 0.9

PR #427, opened 2015-08-14
Fix path for Python test

Bonczidai:


Bonczidai:
backslashes in file path need to be escaped so portions of the path like '\b' will not be interpreted as special characters

orius123:
Didn't happen for me on Linux :smile: 

Can we do it in the cli inputs as well?

PR #428, opened 2015-08-14
Sporadic failure travis

### LINKED ISSUE #365: Sporadic test failures - Travis
SimpleFlowTest.testFlowWithGlobalSession
DataFlowTest.testBindingsFlow

detailed log here: https://gist.github.com/Bonczidai/864f936c36238b5c903b

Bonczidai:
fixes #365


Bonczidai:
I think the problem is that events get lost sometimes if the `onEvent()` method is not syncronized - restarted the build 4 times, the problem seem to be gone

Bonczidai:
also adding @orius123

orius123:
I don't like to solve stuff without understanding..
As I see it, each test start after the former ended so events should not be mixed..
Will merge, but I hope we are not hiding a bug :bug: here.
Consider this as a future "Told you so"

PR #431, opened 2015-08-16
update score version to 0.2-rc2

meirwah:
update score version to 0.2-rc2


orius123:
How can we run content test with it?

meirwah:
I do not follow ... what do you mean?
isn't the tests part of the release process?

PR #438, opened 2015-08-25
Fix the inputs binding event to keep the order of the inputs

### LINKED ISSUE #357: The order of inputs/outputs during binding events is random, and should be as it defined in the flow itself

oritstone:
Fixes #357 


Bonczidai:
I'd also add tests to verify the inputs / outputs order.. WDYT? @oritstone @meirwah

meirwah:
agree @Bonczidai

Bonczidai:
added tests in `runtime` module.. please review @meirwah

meirwah:
:+1:

PR #443, opened 2015-09-02
use spring-io platform third party recommendations and also use latest score

i022098:


orius123:
Y did U change Jcommander version?

i022098:
yes, i changed it to fit spring-io versions

orius123:
Well. 1.35 was released two years ago. So just make sure there are no important features/bug-fixes that will hurt us.

oritstone:
I went over all their commits since 1.35
It seems that there were no bug fixes, only added 3 new small features that we do not use
@orius123 do you see a problem with this version downgrade?

orius123:
If U went over it than no.

PR #445, opened 2015-09-08
Changed ssh version

tudorlesan:
Signed-off-by: Lesan Tudor tudor-andrei.lesan@hp.com


tudorlesan:
should fix CloudSlang/cloud-slang-content#318

PR #448, opened 2015-09-14
Update ssh action version

Bonczidai:
SSH fixes / enhancements:
- reduce polling interval (will be huge improvement for build time)
- session is down
- populate exception output in case of timeout


meirwah:
:+1:

PR #466, opened 2015-10-12
Add Python 'get' function

### LINKED ISSUE #420: Introduce get('key', 'default') function for outputs [1]
Many times when binding outputs the user need to check if a certain key is available in the context and if not bind a default value.
The way it is usually done is by `key' if 'key' in locals() else 'default'`.
I suggest we put a function that does this exact logic in the context when binding outputs,
than the user will be able to simplify his outputs to `get('key', 'default')`

Bonczidai:
fixes #420 

Signed-off-by: Bonczidai Levente levente.bonczidai@hpe.com

### DIFF
--- a/cloudslang-runtime/src/main/java/io/cloudslang/lang/runtime/bindings/ScriptEvaluator.java
+++ b/cloudslang-runtime/src/main/java/io/cloudslang/lang/runtime/bindings/ScriptEvaluator.java
@@ -13,41 +13,47 @@ package io.cloudslang.lang.runtime.bindings;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Component;
 
-import javax.script.ScriptContext;
-import javax.script.ScriptEngine;
-import javax.script.ScriptException;
-import javax.script.SimpleScriptContext;
+import javax.script.*;
 import java.io.Serializable;
 import java.util.Map;
 
 /**
  * @author stoneo
- * @since 06/11/2014
  * @version $Id$
+ * @since 06/11/2014
  */
 @Component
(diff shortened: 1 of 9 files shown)


Bonczidai:
clarify:

`get('key', 'default_value')` : 

- this Python function can be used in any place where we evaluate a Python expression (outputs, inputs, loop expressions) but not in Python actions
- returns the `value` associated with `key` if the `key` is defined and its `value` is not **null** else `default value`

oritstone:
Nice work :+1: 
Maybe you should also add tests for:
- Using the new get() in operation result evaluation
- Add to the flow an input that uses the actual python locals().get() function to see it still works

Another comment: 
should it be called get() as is the original python action? Maybe we should find another name? WDYT? @Bonczidai @mihaitusa @gituser173 @orius123 @meirwah @avigail-oron

orius123:
I'm fine with the name.

orius123:
What is the performance impact? Cant we declare it once as a static/utility method somehow?

Bonczidai:
@orius123 I wouldn't say the performance impact is notable (it adds 1 ms to each evaluation) 

I've investigated adding it as a utility function, but seems like `ScriptEngine` and `Jython` does not support such a thing.. neither helped the hacking.. main problem is, when the `get` function is evaluated it saves the current context as globals so we cannot save the function as a `PyFunction` for example (doesn't take the values from later script context)

p.s. I am also fine with the name

oritstone:
@SamMarkowitz we need to add it to our docs as well

SamMarkowitz:
Yes, I will add and issue on the docs repo.

PR #479, opened 2015-11-10
Fix cli log: missing call arguments for python action

### LINKED ISSUE #331: CLI log: Missing CALL_ARGUMENTS for Python action [1]
Observed the following in CLI log:

Java action - call arguments logged:
```
STEP_TYPE=ACTION, STEP_NAME=null, TYPE=EVENT_ACTION_START, CALL_ARGUMENTS={characterSet=UTF-8, agentForwarding=null, command=docker ps -a, timeout=600000, password=null, pty=false, privateKeyFile=c:/.../id_rsa, port=22, host=111.111.111.111, arguments=null, closeSession=false, username=core}
```

Python action - call arguments not logged:
```
STEP_TYPE=ACTION, STEP_NAME=null, TYPE=EVENT_ACTION_START, CALL_ARGUMENTS={}
```

however there are inputs declared in the operation.

Shouldn't call arguments appear in the log also for Python actions (EVENT_ACTION_START events)?

maayan-avraham:
Fixes #331
Signed-off-by: Maayan Avraham <maayan.avraham@hpe.com>

### DIFF
--- a/cloudslang-runtime/src/main/java/io/cloudslang/lang/runtime/steps/ActionSteps.java
+++ b/cloudslang-runtime/src/main/java/io/cloudslang/lang/runtime/steps/ActionSteps.java
@@ -63,7 +63,7 @@ public class ActionSteps extends AbstractSteps {
         Map<String, Serializable> callArgumentsDeepCopy = new HashMap<>();
 
         for (Map.Entry<String, Serializable> entry : callArguments.entrySet()) {
-            callArgumentsDeepCopy.put(new String(entry.getKey().toString()),  SerializationUtils.clone(entry.getValue()));
+            callArgumentsDeepCopy.put(entry.getKey(),  SerializationUtils.clone(entry.getValue()));
         }
 
         Map<String, SerializableSessionObject> serializableSessionData = runEnv.getSerializableDataMap();
@@ -316,6 +316,4 @@ public class ActionSteps extends AbstractSteps {
     private void cleanInterpreter(PythonInterpreter interpreter) {
         interpreter.setLocals(new PyStringMap());
     }
-
-
 }
--- a/cloudslang-runtime/src/main/java/io/cloudslang/lang/runtime/steps/ActionSteps.java
+++ b/cloudslang-runtime/src/main/java/io/cloudslang/lang/runtime/steps/ActionSteps.java
@@ -17,11 +17,12 @@ import io.cloudslang.lang.entities.ScoreLangConstants;
 import io.cloudslang.lang.runtime.env.ReturnValues;
 import io.cloudslang.lang.runtime.env.RunEnvironment;
 import io.cloudslang.lang.runtime.events.LanguageEventData;
-import org.apache.commons.lang.StringUtils;
-import org.apache.commons.lang3.tuple.Pair;
-import org.apache.log4j.Logger;
 import io.cloudslang.score.api.execution.ExecutionParametersConsts;
 import io.cloudslang.score.lang.ExecutionRuntimeServices;
+import org.apache.commons.lang.StringUtils;
+import org.apache.commons.lang3.SerializationUtils;
+import org.apache.commons.lang3.tuple.Pair;
+import org.apache.log4j.Logger;
 import org.python.core.*;
 import org.python.util.PythonInterpreter;
 import org.springframework.beans.factory.annotation.Autowired;
@@ -30,11 +31,7 @@ import org.springframework.stereotype.Component;
 import java.io.Serializable;
 import java.lang.annotation.Annotation;
 import java.lang.reflect.Method;
-import java.util.ArrayList;
-import java.util.HashMap;
-import java.util.Iterator;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
 
 import static io.cloudslang.score.api.execution.ExecutionParametersConsts.EXECUTION_RUNTIME_SERVICES;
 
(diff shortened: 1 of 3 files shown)


cloudslang-runtime/src/main/java/io/cloudslang/lang/runtime/steps/ActionSteps.java
```diff
@@ -63,10 +60,16 @@ public void doAction(@Param(ScoreLangConstants.RUN_ENV) RunEnvironment runEnv,
 
         Map<String, Serializable> returnValue = new HashMap<>();
         Map<String, Serializable> callArguments = runEnv.removeCallArguments();
+        Map<String, Serializable> callArgumentsDeepCopy = new HashMap<>();
+
+        for (Map.Entry<String, Serializable> entry : callArguments.entrySet()) {
+            callArgumentsDeepCopy.put(new String(entry.getKey().toString()),  SerializationUtils.clone(entry.getValue()));
```
Bonczidai: you can use simply `entry.getKey()` here instead of `new String(entry.getKey().toString()` because of diamond operator (`Map.Entry<String, Serializable> entry`)

cloudslang-runtime/src/main/java/io/cloudslang/lang/runtime/steps/ActionSteps.java
```diff
@@ -313,4 +316,6 @@ private void executePythonScript(PythonInterpreter interpreter, String script, M
     private void cleanInterpreter(PythonInterpreter interpreter) {
         interpreter.setLocals(new PyStringMap());
     }
+
+
```
Bonczidai: how about only one newline?

cloudslang-runtime/src/test/java/io/cloudslang/lang/runtime/steps/ActionStepsTest.java
```diff
@@ -135,6 +136,75 @@ public void doJavaActionWrongClassTest() {
...
+        Collection<ScoreEvent> events = runtimeServices.getEvents();
+
+        Assert.assertFalse(events.isEmpty());
+        ScoreEvent eventActionStart = null;
+        for(ScoreEvent event:events){
+            if(event.getEventType().equals(ScoreLangConstants.EVENT_ACTION_START)){
+                eventActionStart = event;
+            }
+        }
+        assert eventActionStart != null;
```
Bonczidai: I am not much of an `assert` fan in unit tests -- I think you should use `assertNotNull` instead

cloudslang-runtime/src/test/java/io/cloudslang/lang/runtime/steps/ActionStepsTest.java
```diff
@@ -135,6 +136,75 @@ public void doJavaActionWrongClassTest() {
...
+        //invoke doAction
+        actionSteps.doAction(runEnv,new HashMap<String, Object>(), PYTHON, "", "", runtimeServices, userPythonScript, 2L, "operationName");
+
+        Collection<ScoreEvent> events = runtimeServices.getEvents();
+
+        Assert.assertFalse(events.isEmpty());
+        ScoreEvent eventActionStart = null;
+        for(ScoreEvent event:events){
+            if(event.getEventType().equals(ScoreLangConstants.EVENT_ACTION_START)){
+                eventActionStart = event;
```
Bonczidai: with this logic you verify that the event was fired at least once, maybe you should check that it was fired exactly once

cloudslang-runtime/src/test/java/io/cloudslang/lang/runtime/steps/ActionStepsTest.java
```diff
@@ -135,6 +136,75 @@ public void doJavaActionWrongClassTest() {
...
+            }
+        }
+        assert eventActionStart != null;
+        Map<String, Serializable> data = (HashMap)eventActionStart.getData();
+        for (Map.Entry<String, Serializable> entry : data.entrySet()) {
```
Bonczidai: how I would refactor this section (e.g. no need for `for`):

- cast `.getData()` to `LanguageEventData`
- update `LanguageEventData` with getter / setter for `CALL_ARGUMENT` const
- use this getter to retrieve the call args
- verify call args content is equal to your inputs (`callArguments.put("index", 1);`)
- don't `return` and `Assert.fail();` you can simply use `assertEquals("message", expected, actual)`

cloudslang-runtime/src/test/java/io/cloudslang/lang/runtime/steps/ActionStepsTest.java
```diff
@@ -135,6 +136,75 @@ public void doJavaActionWrongClassTest() {
...
+            if (entry.getKey().equals("CALL_ARGUMENTS")){
+                Assert.assertNotNull(entry.getValue());
+                return;
+            }
+        }
+        Assert.fail();
+    }
+
+    @Test
+    public void doActionJavaActionCheckCallArgumentsOnEvent() {
```
Bonczidai: same comments apply as in python method case

Bonczidai:
really nice fix - me like it

cloudslang-runtime/src/test/java/io/cloudslang/lang/runtime/steps/ActionStepsTest.java
maayan-avraham: @Bonczidai This test is only for checking that the call arguments are correct, so why should I verify here that the event fired exactly once?

Bonczidai: ok -- how about adding a `break` so we will not iterate over the remaining part of the collection?

maayan-avraham: I agree with you, thanks

maayan-avraham:
@Bonczidai  it's ready for merge

PR #517, opened 2015-12-09
update http-client-version

gituser173:


Bonczidai:
wrong issue number in first comment (I don't think we need any `fixes` here)
title should be something like - update http-client-version, right?
branch name also would be better like: update-http-client-version WDYT?

@gituser173, please fix the first 2

Bonczidai:
:+1:

PR #521, opened 2015-12-14
Add toString, equals, hashcode methods to entities

### LINKED ISSUE #510: Entity references in logs
some entities still appear as obj refs (e.g. `io.cloudslang.lang.entities.ResultNavigation@5b7d`) instead of toString representation

Bonczidai:
fixes #510 

they will also be logged properly

### DIFF
--- a/cloudslang-entities/src/main/java/io/cloudslang/lang/entities/AsyncLoopStatement.java
+++ b/cloudslang-entities/src/main/java/io/cloudslang/lang/entities/AsyncLoopStatement.java
@@ -10,8 +10,6 @@
 package io.cloudslang.lang.entities;
 
 import org.apache.commons.lang3.Validate;
-import org.apache.commons.lang3.builder.EqualsBuilder;
-import org.apache.commons.lang3.builder.HashCodeBuilder;
 
 import java.io.Serializable;
 
@@ -41,29 +39,6 @@ public class AsyncLoopStatement extends LoopStatement implements Serializable {
 
     public String getVarName() {
         return varName;
-    }
-
-    @Override
-    public boolean equals(Object o) {
-        if (this == o)
-            return true;
-        if (o == null || getClass() != o.getClass())
-            return false;
-
-        AsyncLoopStatement that = (AsyncLoopStatement) o;
-
-        return new EqualsBuilder()
-                .append(varName, that.varName)
-                .append(getExpression(), that.getExpression())
-                .isEquals();
-    }
-
-    @Override
-    public int hashCode() {
-        return new HashCodeBuilder()
-                .append(varName)
-                .append(getExpression())
-                .toHashCode();
     }
 
 }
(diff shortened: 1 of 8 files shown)


cloudslang-entities/src/main/java/io/cloudslang/lang/entities/AsyncLoopStatement.java
```diff
@@ -43,27 +41,4 @@ public String getVarName() {
         return varName;
     }
 
-    @Override
```
oritstone: Why did you move the toString(), equals() & hashcode() methods to the parent class? 
You miss the extending class's specific fields, such as `varName` in this case

cloudslang-entities/src/main/java/io/cloudslang/lang/entities/LoopStatement.java
```diff
@@ -37,4 +40,20 @@ protected LoopStatement() {
     public String getExpression() {
         return expression;
     }
+
```
oritstone: Why do we use these libraries? they use reflection to determine the fields of each object. It is probably slower than doing it manually.

cloudslang-entities/src/main/java/io/cloudslang/lang/entities/AsyncLoopStatement.java
Bonczidai: subclass fields are included - we only the method in superclass

cloudslang-entities/src/main/java/io/cloudslang/lang/entities/LoopStatement.java
Bonczidai: I prefer these methods over doing it manually, since
- we don't need to define each method in each subclass
- reduced code size
- standard output format
- can be considered best practice, better control over data

I like this approach, but we can go with explicit methods for each subclass - WDYT?

oritstone: I think that the elegance and code size is not enough in this case...
We will loose in the performance area for each time we need to compare 2 objects or toString() them, and I think it will probably happen a lot...

Bonczidai:
@oritstone, how about this style? ->

```java
    @Override
    public String toString() {
        return "CompilationArtifact{" +
                "executionPlan=" + executionPlan +
                ", dependencies=" + dependencies +
                ", inputs=" + inputs +
                ", systemProperties=" + systemProperties +
                '}';
    }

    @Override
    public boolean equals(Object o) {
        if (this == o) return true;

        if (o == null || getClass() != o.getClass()) return false;

        CompilationArtifact that = (CompilationArtifact) o;

        return new EqualsBuilder()
                .append(executionPlan, that.executionPlan)
                .append(dependencies, that.dependencies)
                .append(inputs, that.inputs)
                .append(systemProperties, that.systemProperties)
                .isEquals();
    }

    @Override
    public int hashCode() {
        return new HashCodeBuilder(17, 37)
                .append(executionPlan)
                .append(dependencies)
                .append(inputs)
                .append(systemProperties)
                .toHashCode();
    }
```

oritstone:
Seems OK. We should use standard equals() hashCode() & toString() methods for each class

orius123:
Isn't there any `toString()` builder? For consistency.

Bonczidai:
:+1:  there is, I used it in the commit, just didn't update the comment

orius123:
Great minds..

PR #535, opened 2016-01-07
Fix default folder

### LINKED ISSUE #534: Default dir for system properties and inputs not working
Doc says: 
```
System property files can be loaded automatically if placed in a folder named properties in the directory from which the CLI is run.
```
does not work

Bonczidai:
fixes #534 


Bonczidai:
@SamMarkowitz please take a look on this issue
@gituser173 please review and validate for `inputs` folder, I verified it for system properties

SamMarkowitz:
I tested your fixes and they work for me when the folders are placed inside the bin folder.

But now that's it's sat in my head for a bit, I seem to remember that the functionality was supposed to be that the properties or inputs folder(s) would be in whichever directory the user was in when he ran the CLI. For example, if I was in C:\ and ran something like `C:\> path\to\cli\cslang.bat`, and then ran a flow in the CLI, the folders (properties and/or inputs) would need to be in C:\. 

That is what the docs mean by "in the directory from which the CLI is run" - which obviously wasn't clear enough, because even I couldn't remember what that meant.

SamMarkowitz:
I forgot to mention that I tested the above way on the last release and that didn't work either.

PR #538, opened 2016-01-11
System properties

### LINKED ISSUE #530: enhance system properties
need to re-check system properties behavior:
- do we want to support many in one file?
- dont we want to allow task input to use it?
- do we want to enable using it inside expression vs. know who uses it....

Bonczidai:
fixes #530 


cloudslang-all/src/main/java/io/cloudslang/lang/api/Slang.java
```diff
@@ -40,7 +40,7 @@
 	 * @param systemProperties the system properties for the flow or operation run
 	 * @return the execution ID in score
 	 */
-	public Long run(CompilationArtifact compilationArtifact, Map<String, ? extends Serializable> runInputs, Map<String, ? extends Serializable> systemProperties);
+	public Long run(CompilationArtifact compilationArtifact, Map<String, ? extends Serializable> runInputs, Map<String, String> systemProperties);
```
orius123: SP are only strings?

Bonczidai: SP values will be strings (SP will be objects to support features like encryption)

Bonczidai:
however SPs will be objects (to support sensitive data for example), for now I am leaving them as maps -- better change it when developing SD

Bonczidai:
ready for review (not for merge)

in the meantime, I will:
- add more tests, validation
- CLI command to inspect a properties file

Bonczidai:
insp

SamMarkowitz:
Haven't run into any issues yet with my testing etc., but here are a few observations:

1. imports - this key seems to be legal in both system properties files and operations, but I'm not sure what it would be used for
2. inspect - seems a bit inconsistent to me. When called on other types of .sl files it parses the documentation, but in a system properties files it goes through the actual "code". Is there no plan to have a formal documentation structure for system properties files? Is there no need? 
3. Does it make sense to be able to import a systems properties file into a flow/operation and create an alias so you don't always need to write out the fully qualified name of all the properties?

Bonczidai:
1. I think also `name` is an ignored keyword in properties files - we have such cases - e.g. in flow you can have ignored `operation` or `properties` keywords (if we want to validate a stricter structure, I think we should open a new issue for it)
2. probably - in that case makes sense to create a new command for loading SPs (again I think we should open a new issue for the doc) -- I'll do these
3. good point - but I think we can wait with this, for now the SP names are usually short / manageable

cloudslang-cli/src/main/java/io/cloudslang/lang/cli/SlangCLI.java
```diff
@@ -127,11 +130,32 @@ private boolean invalidVerboseInput(String verbose) {
...
-        return metadataHelper.extractMetadata(executableFile);
+        if (compilerHelper.isExecutable(slangFile)) {
+            return metadataHelper.extractMetadata(new File(slangFile));
+        } else {
+            Set<SystemProperty> systemProperties = compilerHelper.loadSystemProperties(Lists.newArrayList(slangFile));
+            return prittyPrintSystemProperties(systemProperties);
+        }
+    }
+
+    private String prittyPrintSystemProperties(Set<SystemProperty> systemProperties) {
```
oritstone: Should be pretty :)

cloudslang-cli/src/main/java/io/cloudslang/lang/cli/SlangCLI.java
```diff
@@ -127,11 +130,32 @@ private boolean invalidVerboseInput(String verbose) {
...
+    )
     public String inspectExecutable(
-            @CliOption(key = {"", "f", "file"}, mandatory = true, help = PATH_TO_FILENAME_HELP) final File executableFile
+            @CliOption(key = {"", "f", "file"}, mandatory = true, help = PATH_TO_FILENAME_HELP) final String slangFile
     ) throws IOException {
-        return metadataHelper.extractMetadata(executableFile);
+        if (compilerHelper.isExecutable(slangFile)) {
+            return metadataHelper.extractMetadata(new File(slangFile));
+        } else {
+            Set<SystemProperty> systemProperties = compilerHelper.loadSystemProperties(Lists.newArrayList(slangFile));
```
oritstone: Do we support receiving a list of system props files to the CLI?

cloudslang-cli/src/main/java/io/cloudslang/lang/cli/utils/CompilerHelperImpl.java
```diff
@@ -141,16 +131,58 @@ public String convert(File from) {
...
+            } catch(Throwable ex) {
+                String errorMessage = "Error loading file: " + propFile + " nested exception is " + ex.getMessage();
+                logger.error(errorMessage, ex);
+                throw new RuntimeException(errorMessage, ex);
+            }
+        }
+        return result;
+    }
+
+    private List<String> loadDefaultFiles(List<String> files, String[] extensions, String directory, boolean recursive) {
```
oritstone: I think the method should return the File objects instead of returning only the paths and then creating the File again in the calling methods

cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/SlangCompilerImpl.java
```diff
@@ -66,4 +68,43 @@ public Executable preCompile(SlangSource source) {
...
+            return extractProperties(parsedSlang);
+        } catch (Throwable ex) {
+            throw new RuntimeException(
+                    "Error loading properties file: " + source.getName() + " nested exception is " + ex.getMessage(),
+                    ex
+            );
+        }
+    }
+
+    private ParsedSlang parseSource(SlangSource source) {
```
oritstone: This method is not for parsing a general Slang source, it's only for parsing system properties source file. So we should change the name to describe what it does

cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/SlangCompilerImpl.java
```diff
@@ -66,4 +68,43 @@ public Executable preCompile(SlangSource source) {
...
+            throw new RuntimeException("Source: " + parsedSlang.getName() + " is not a valid system property file.");
+        }
+        return parsedSlang;
+    }
+
+    private Set<SystemProperty> extractProperties(ParsedSlang parsedSlang) {
+        String namespace = parsedSlang.getNamespace();
+        Map<String, Object> rawSystemProperties = parsedSlang.getProperties();
+        Set<SystemProperty> properties = new HashSet<>();
+        Set<Map.Entry<String, Object>> entrySet = rawSystemProperties.entrySet();
```
oritstone: Did you check if you call this method with a parsedSlang object that is not system properties that there are no NullPointer exceptions?

cloudslang-compiler/src/test/resources/inputs_with_functions.sl
```diff
@@ -0,0 +1,30 @@
...
+        default: ${get('key', get_sp('a.b.c.key'))}
+    - input8:
+        default: ${get('key', get_sp('a.b.c.key')) + get_sp('d.e.f.key')}
+    - input9:
+        default: ${get('key', get_sp('a.b.c.key')) + get_sp('a.b.c.key')}
+    - input10:
+        default: ${get_sp(expression)}
+    - input11:
+        default: ${get_sp( 'a.b.c.key'     )}
+  action:
```
oritstone: What about expressions that contain 2 calls for the same functions such as:
- ${ get(key, default) + get(key2, default) }
- ${ get_sp(key, default) + get_sp(key2, default) }

cloudslang-runtime/src/main/java/io/cloudslang/lang/runtime/bindings/scripts/ScriptEvaluator.java
```diff
@@ -0,0 +1,122 @@
...
+                    interpreter.set(SYSTEM_PROPERTIES_MAP, prepareSystemProperties(systemProperties));
+                    functions += GET_SP_FUNCTION_DEFINITION;
+                    functions = appendDelimiterBetweenFunctions(functions);
+                    break;
+            }
+        }
+
+        if (StringUtils.isNotEmpty(functions)) {
+            functions = trimScript(functions);
+            exec(interpreter, functions);
```
oritstone: Why do we execute the script? Is that the way to define the new functions on the interpreter?

oritstone:
Very good work, Levi :+1: 

One comment though - I think that the system properties should also be a member on the Executable. 
The *preCompile()* API returns the Executable, and we should have all the system properties that this executable uses (not recursively, only direct)
The *CompilationArtifact* should calculate the system properties recursively.
This is needed for validations for example, similar for the need for dependencies of the Executable.

I suggest that you merge these changes after fixing my small comments, and then we can discuss further how to address this issue

cloudslang-cli/src/main/java/io/cloudslang/lang/cli/SlangCLI.java
Bonczidai: I think we can go with one file for now, after syncing with @SamMarkowitz , we agreed to add a separate `list properties.sl` cmd for displaying SPs from a file

cloudslang-runtime/src/main/java/io/cloudslang/lang/runtime/bindings/scripts/ScriptEvaluator.java
Bonczidai: exactly - that is how the interpreter knows about the functions with the current context

Bonczidai:
for @SamMarkowitz :

er_message

list

Bonczidai:
thanks @oritstone for the review, I'll created #549 for the SPs on Executable issue

PR #544, opened 2016-01-20
Flow description - NullPointerException when section missing

### LINKED ISSUE #543: Flow description - NullPointerException when section missing
Some sections don't always need to be in the documentation. If there are no outputs or if the results are just the defaults, those sections don't need to be there.

If a section is missing, the inspect gives:

```
Command failed java.lang.RuntimeException: java.lang.NullPointerException
```

Example - works fine:

```
## description: operation_description
##
## inputs:
##   - input_1: input_description
## outputs:
##   - output_1: output_description
## results:
##   - SUCCESS: good
``` 

Example - throws error:

```
## description: operation_description
##
## inputs:
##   - input_1: input_description
## outputs:
##   - output_1: output_description
``` 

Also, if one of the double hashes is just a single hash you get the same error which isn't so descriptive.

gituser173:
Fixes #543 


Bonczidai:
not only outputs and results but also inputs are optional in description section IMHO, since we can have no inputs for a flow -- please test all three cases

preferable create a folder ~ `metadata` for you .sl sources (if you have many)

Bonczidai:
.. and please test this also: " if one of the double hashes is just a single hash you get the same error which isn't so descriptive."

PR #548, opened 2016-02-01
Flow description similar to javadoc

### LINKED ISSUE #545: Flow description similar to javadoc
After internal discussions we reached the conclusion that we have two options:
First would be to mark every line with two characters "#!" so we know which lines contain the description. 
```yaml
#   (c) Copyright 2016 Hewlett-Packard Development Company, L.P.
#   All rights reserved. This program and the accompanying materials
#   are made available under the terms of the Apache License v2.0 which accompany this distribution.
#
#   The Apache License is available at
#   http://www.apache.org/licenses/LICENSE-2.0
#
#! @description: general desc
#! @input input1_name: input1 desc
#! @input input2_name: input2 description
#!                     optional
#!                     default: default value
#!                     valid: valid values
#!                     example: example value
#! @output output1_name: output1 desc
#! @result result1_name: result1 desription
#! @result result2_name: result2 description

imports:
  containers: io.cloudslang.docker.containers

flow:
  name: create_cluster
```
The second option would be to mark the beginning and the ending of the description block with "#!" or with some other character combination.
 ```yaml
#   (c) Copyright 2016 Hewlett-Packard Development Company, L.P.
#   All rights reserved. This program and the accompanying materials
#   are made available under the terms of the Apache License v2.0 which accompany this distribution.
#
#   The Apache License is available at
#   http://www.apache.org/licenses/LICENSE-2.0
#
#! 
# @description: general desc
# @input input1_name: input1 desc
# @input input2_name: input2 description
#                     optional
#                     default: default value
#                     valid: valid values
#                     example: example value
# @output output1_name: output1 desc
# @result result1_name: result1 desription
# @result result2_name: result2 description
#!

imports:
  containers: io.cloudslang.docker.containers

flow:
  name: create_cluster
```
@CloudSlang/slang-functional-reviewers please give your feedback

### PR DESCRIPTION
**Pull request by gituser173**
Fixes #545

### CODE REVIEW

File: cloudslang-cli/src/main/java/io/cloudslang/lang/cli/utils/MetadataHelperImpl.java [hunk 1]
```diff
@@ -63,8 +69,28 @@ private Metadata cleanWhiteSpaces(Metadata metadata, String lineBreak) {
     }
 
     private void cleanMapWhiteSpaces(Map<String, String> map, String lineBreak) {
-        for (String key : map.keySet()) {
-            map.put(key, map.get(key).trim().replaceAll("[\\s]+" + lineBreak, lineBreak));
+        for (Map.Entry<String, String> entry : map.entrySet()) {
+            map.put(entry.getKey(), entry.getValue().trim().replaceAll("[\\s]+" + lineBreak, lineBreak));
```
Bonczidai: is it possible for the value to be null?

better use StringUtils or check yourself

cloudslang-cli/src/main/java/io/cloudslang/lang/cli/utils/MetadataHelperImpl.java
```diff
@@ -63,8 +69,28 @@ private Metadata cleanWhiteSpaces(Metadata metadata, String lineBreak) {
...
+        protected Set<Property> getProperties(Class<? extends Object> type)
+                throws IntrospectionException {
+            Set<Property> result = new LinkedHashSet<>();
+            Set<Property> set = super.getProperties(type);
+            for (Property property : set) {
+                if (property.getType().equals(String.class)) {
+                    result.add(property);
+                }
+            }
+            for (Property property : set) {
```
Bonczidai: can you extract a generic method for the `for` or use only one loop and two `if`s?
what happens when the type is not `String` or `Map`? should those values be ignored?

cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/parser/MetadataParser.java
```diff
@@ -28,40 +28,109 @@
...
+    }
+
+    private DescriptionTag getContainedTag(String line) {
+        for (DescriptionTag descriptionTag : DESCRIPTION_TAGS_LIST) {
+            if (line.contains(descriptionTag.getValue())) return descriptionTag;
+        }
+        return null;
+    }
+
+    private boolean lineContainsSomeTag(String line) {
```
Bonczidai: can we skip `some`?

cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/parser/MetadataParser.java
```diff
@@ -28,40 +28,109 @@
...
+            return fullMap;
         } catch (Throwable e) {
-            throw new RuntimeException("There was a problem parsing the description in YAML format: " +
+            throw new RuntimeException("There was a problem parsing the description: " +
                     source.getName() + ".\n" + parserExceptionHandler.getErrorMessage(e), e);
         }
     }
 
-    private String extractDescriptionYAMLString(SlangSource source) {
+    private void checkMapOrder(Map<String, String> fullMap) {
```
Bonczidai: do you need this method at runtime? why not adding a test for it?

Bonczidai:
insepect - CLI:
- when flow have no description section:
```
Command failed java.lang.RuntimeException: java.lang.RuntimeException: There was a problem parsing the description: build_image.sl.
null
```
-> ~ no metadata found

- when
```
#!! @description: Builds a Docker image based on a Dockerfile.
#!!#
```
->
```
Command failed java.lang.RuntimeException: java.lang.IllegalArgumentException: Operation/Flow must have a description
```

- when
```
#!!
#1 @description: Builds a Docker image based on a Dockerfile.
```
->
```
Command failed java.lang.RuntimeException: java.lang.IllegalArgumentException: Operation/Flow must have a description
```
- it passes when no closing tag, is this the desired behavior?
- when part of the description is after the root tag (flpw / op) it passes, is this the desired behavior?
- when there is white space before the `#1` markers, it fails - is this the desired behavior?

can you take a look whether we can refine these error messages?

cloudslang-cli/src/main/java/io/cloudslang/lang/cli/utils/MetadataHelperImpl.java
gituser173: All the `String`'s are being added to the result, only after that all the `Map`'s are added to the result. The properties are ordered by name(prerequisites is after outputs). The Metadata class only contains `String` and `Map` properties(this is a custom order specific to the Metadata class). A new method can be extracted for those two `for` loops.

cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/parser/MetadataParser.java
gituser173: replaced it with `a`

cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/parser/MetadataParser.java
gituser173: don't we want the users to keep an order in the description? @description, @prerequisites(optional), @input(optional), @output(optional), @result(optional)
do we want to let users mix inputs with outputs etc.?

cloudslang-cli/src/main/java/io/cloudslang/lang/cli/utils/MetadataHelperImpl.java
gituser173: added unit test for this scenario

gituser173:
@Bonczidai after discussions we decided that values on the same lines as the starting tag and the closing tag should not be supported.

Bonczidai:
:ok_hand: 

two more comments:
- please resolve conflicts with master
- I'd suggest to use short descriptive commit messages in the future :)

Issue #549, opened 2016-02-03
Properties files with .sl extension


Bonczidai:
rethink how to distinguish between SP and executable .sl files (now it's in CLI / builder naive implementation)

Bonczidai:
the issue that we need to discuss: should SPs be really .sl files?

SP files have a certain structure, but it does not necessarily need to be CloudSlang file. `.sl` files were only executables (flows / ops) before. We need to distinguish clearly between executables / SP files (in the future, maybe inputs files).

[CloudSlang API](https://github.com/CloudSlang/cloud-slang/blob/master/cloudslang-all/src/main/java/io/cloudslang/lang/api/Slang.java) has methods like:
```Java
CompilationArtifact compile(SlangSource source, Set<SlangSource> dependencies);
Long run(CompilationArtifact compilationArtifact, Map<String, ? extends Serializable> runInputs, Set<SystemProperty> systemProperties);
Long compileAndRun(SlangSource source, Set<SlangSource> dependencies, Map<String, ? extends Serializable> runInputs, Set<SystemProperty> systemProperties);
Set<SystemProperty> loadSystemProperties(SlangSource source);
```

the api's methods accept a source and do something with it, at this level it is expected from the caller to know the type (now we distinguish in upper layers like CLI based on top level tag (e.g. properties / flow / operation - not so cool)

so I see two options here:
- we really want SP files to be `.sl` -- lot of redesign, common class for Executable and SPs, SP files will be **compileable** (no loadSP() methods)
- they will have another extension (`.properties` / `.prop`/ `.sp` maybe `.properties.sl`) - they will be loadable, we know in upper layers clearly what to pass to compile and what to pass to loadSP methods

I don't say that the `.sl` approach is bad, just too much effort comparing to the benefits -- logically SP files have structure which is verified, but I think they should remain loadable only (not compilable)

Bonczidai:
@oritstone @meshipeer let's hear WDYT

orius123:
what are the benefits? namespace?
can't it be just YAML?

oritstone:
I think that it is better that system properties files will have a specific, different extension (i.e. `.sp`, `spl`)
This way we can distinguish between executable & system properties and still be able to have a pre-defined structure for system properties and use namespace

Bonczidai:
I would also go with `.sp`

are there any objections?

meshipeer:
[...]

PR #550, opened 2016-02-04
Fix - filter SP files for builder

Bonczidai:


Bonczidai:
maybe we need to rethink how we distinguish between SP and executable .sl files - mentioned it in #549

PR #553, opened 2016-02-08
System properties

### LINKED ISSUE #549: Properties files with .sl extension

### LINKED ISSUE #556: System Property leftovers
- System Property dependencies should be stored on Executable level

The system properties should also be a member on the Executable. 
The preCompile() API returns the Executable, and we should have all the system properties that this executable uses (not recursively, only direct)
The CompilationArtifact should calculate the system properties recursively.
This is needed for validations for example, similar for the need for dependencies of the Executable.

- replace static method with constructor for SPs + empty namespace

Bonczidai:
fixes #556

### DIFF
--- a/cloudslang-all/src/test/java/io/cloudslang/lang/api/SlangImplTest.java
+++ b/cloudslang-all/src/test/java/io/cloudslang/lang/api/SlangImplTest.java
@@ -147,7 +147,7 @@ public class SlangImplTest {
 
     @Test
     public void testRun(){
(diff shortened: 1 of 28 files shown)


Bonczidai:
@oritstone please review - this PR fixes the following issues:

- System Property dependencies should be stored on Executable level
- replace static method with constructor for SPs + empty namespace

the SP extension leftover will be fixed on another PR

oritstone:
Nice work :+1:

PR #575, opened 2016-02-26
Compiler error handling

### LINKED ISSUE #564: Be able to extract language data from flows that may be incorrect but valid Yaml
When creating new flow the workflow is empty.
When trying to compile this flow in order to get the namespace, for example, we get an error that our workflow is empty.
Can we make such scenarios legal in compiler?

thank

Bonczidai:
fixes #564 

### DIFF
--- a/cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/SlangCompiler.java
+++ b/cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/SlangCompiler.java
@@ -26,7 +26,8 @@ public interface SlangCompiler {
     CompilationArtifact compile(SlangSource source, Set<SlangSource> path);
 
     /**
-     * Pre-compile a CloudSlang source into an {@link io.cloudslang.lang.compiler.modeller.model.Executable}
+     * Pre-compile a CloudSlang source into an {@link io.cloudslang.lang.compiler.modeller.model.Executable}.
+     * If an error is found, an exception is thrown
      * @param source the {@link SlangSource}
      * @return an {@link io.cloudslang.lang.compiler.modeller.model.Executable} object, containing either a flow or
      * operation in the file.
@@ -34,10 +35,11 @@ public interface SlangCompiler {
     Executable preCompile(SlangSource source);
 
     /**
-     * Pre-compile a CloudSlang source into an {@link ExecutableModellingResult}
+     * Pre-compile a CloudSlang source into an {@link ExecutableModellingResult}.
+     * All errors that are found during pre-compilation are collected (an exception is not thrown)
      * @param source the {@link SlangSource}
-     * @return an {@link ExecutableModellingResult} object, containing an executable
-     * which is either a flow or an operations in the file.
+     * @return an {@link ExecutableModellingResult} object, containing an executable which is either a flow
+     * or an operations in the file, and a list of all the errors that were found (if any).
      */
     ExecutableModellingResult preCompileSource(SlangSource source);
 
--- a/cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/SlangCompiler.java
+++ b/cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/SlangCompiler.java
@@ -8,11 +8,12 @@
  */
 package io.cloudslang.lang.compiler;
 
-import java.util.Set;
-
 import io.cloudslang.lang.compiler.modeller.model.Executable;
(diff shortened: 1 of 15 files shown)


cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/modeller/ExecutableBuilder.java
```diff
@@ -91,19 +103,33 @@ public void initScopedTransformersAndKeys() {
         return filter(having(on(Transformer.class).getScopes().contains(scope)), transformers);
     }
 
-    public Executable transformToExecutable(ParsedSlang parsedSlang, String execName, Map<String, Object> executableRawData) {
+    @SuppressWarnings("unchecked")
```
Bonczidai: I am not much of a fan of `SupressWarnings` on methods..

can we use some tricks? like:
```java
@SuppressWarnings("unchecked")
List<Map<String, Map<String, Object>>> onFailureDataHelper = (List<Map<String, Map<String, Object>>>)taskData.values().iterator().next();
onFailureData = onFailureDataHelper;
```

cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/modeller/ExecutableBuilder.java
```diff
@@ -272,45 +308,53 @@ private Workflow compileWorkFlow(List<Map<String, Map<String, Object>>> workFlow
...
-        transformersHandler.validateKeyWords(taskName, taskRawData, ListUtils.union(preTaskTransformers, postTaskTransformers), TaskAdditionalKeyWords, null);
+        errors.addAll(transformersHandler.checkKeyWords(taskName, taskRawData, ListUtils.union(preTaskTransformers, postTaskTransformers), TaskAdditionalKeyWords, null));
 
-        try {
-            preTaskData.putAll(transformersHandler.runTransformers(taskRawData, preTaskTransformers));
-            postTaskData.putAll(transformersHandler.runTransformers(taskRawData, postTaskTransformers));
-        } catch (Exception ex){
-            throw new RuntimeException("For task: " + taskName + " syntax is illegal.\n" + ex.getMessage(), ex);
-        }
+        String errorMessagePrefix = "For task: " + taskName + " syntax is illegal.\n";
```
Bonczidai: I see that you specify a prefix for tasks but you don't do that for executable and action - Y?

Bonczidai:
@ifatgv / @ifatgv4 this looks :100: 

please add lots of test cases to make sure we don't miss anything

cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/SlangCompiler.java
```diff
@@ -27,10 +28,19 @@
...
      * Pre-compile a CloudSlang source into an {@link io.cloudslang.lang.compiler.modeller.model.Executable}
      * @param source the {@link SlangSource}
-     * @return an {@link io.cloudslang.lang.compiler.modeller.model.Executable} object, containing either a flow or operation in the file.
+     * @return an {@link io.cloudslang.lang.compiler.modeller.model.Executable} object, containing either a flow or
+     * operation in the file.
      */
     Executable preCompile(SlangSource source);
 
+    /**
+     * Pre-compile a CloudSlang source into an {@link ExecutableModellingResult}
```
Bonczidai: I think we should mention in the javadoc that this version will collect the exceptions instead of failing

cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/modeller/ExecutableBuilder.java
ifatgv4: Done

cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/modeller/ExecutableBuilder.java
ifatgv4: I tried to keep the error messages as is. Only task errors had a prefix, so I only added it to them.
However, there is no reason not to add such a prefix.
I added one for action and executable

cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/SlangCompiler.java
ifatgv4: Done

Issue #423, opened 2015-08-13
User supplied config file for CLI


Bonczidai:
We could add an optional configuration file (user-supplied) to the CLI with global config data like
default classpath locations, env variables, system properties, quiet mode, debug mode ?

WDYT?

orius123:
like :+1:

meirwah:
:+1:

orius123:
we can add a `--config` flag to specify different file than default

Bonczidai:
started in #576 - later we can also add capabilities of defining default folder for inputs and sp files

PR #576, opened 2016-02-26
Add possibility of config file to CLI

### LINKED ISSUE #485: Add log4j file to cslang-cli.zip
Add "log4j.properties" (under cloudslang-cli - >resources ) to cslang-cli.zip so the user would be able  to configure the file:
1. add it as a task to the build_cli_flow.sl
2. configure spring shell to the new location of log4j file

# suggestion for solution:
*option 1:* create "resources" folder in the zip which will contain log4j file. spring shell supposed to take log4j file from "resources" folder by default.
*option 2:* explicitly mentioning the location of log4j by adding it to "application.properties" file:
logging.config= # Location of the logging configuration file. For instance `classpath:logback.xml` for Logback

(we can check how this is implemented in other applications- maybe in OO and decide)

links:
http://docs.spring.io/spring-boot/docs/current/reference/html/common-application-properties.html

http://docs.spring.io/spring-boot/docs/current/reference/htmlsingle/#boot-features-logging

Bonczidai:
fixes #485 
fixes partially #423  

### DIFF
--- a/cloudslang-cli/src/main/java/io/cloudslang/lang/cli/SlangBootstrap.java
+++ b/cloudslang-cli/src/main/java/io/cloudslang/lang/cli/SlangBootstrap.java
@@ -82,10 +82,10 @@ public class SlangBootstrap {
     private static void configureLogging(Properties userProperties) {
         String filePath = userProperties.getProperty(LOG4J_CONFIGURATION);
         if (StringUtils.isNotEmpty(filePath)) {
-            // log4j will not search for config files in classpath resources
-            System.setProperty("log4j.configuration", filePath);
             // load log4j properties from file
             PropertyConfigurator.configure(filePath);
+            // log4j will not search for config files in classpath resources
+            System.setProperty("log4j.configuration", filePath);
         }
     }
 
--- a/cloudslang-cli/src/main/java/io/cloudslang/lang/cli/SlangBootstrap.java
+++ b/cloudslang-cli/src/main/java/io/cloudslang/lang/cli/SlangBootstrap.java
@@ -15,10 +15,10 @@ import org.apache.commons.lang3.StringUtils;
 import org.springframework.shell.Bootstrap;
 
 import java.io.*;
-import java.util.Enumeration;
 import java.util.HashSet;
 import java.util.Properties;
 import java.util.Set;
+import java.util.Map;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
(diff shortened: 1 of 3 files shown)


cloudslang-cli/src/main/java/io/cloudslang/lang/cli/SlangBootstrap.java
```diff
@@ -21,9 +28,50 @@
...
         System.out.println("Loading..");
         Bootstrap.main(args);
     }
 
+    private static void loadUserProperties() throws IOException {
+        String appHome = System.getProperty("app.home");
+        String propertyFilePath = appHome + File.separator + USER_CONFIG_FILEPATH;
+        File propertyFile = new File(propertyFilePath);
+        Properties rawProperties = new Properties();
+        if (propertyFile.exists()) {
```
gituser173: you could use isFile() here, checks if exists and if it is a file

cloudslang-cli/src/main/java/io/cloudslang/lang/cli/SlangBootstrap.java
```diff
@@ -21,9 +28,50 @@
...
+            try (InputStream propertiesStream = new FileInputStream(propertyFilePath)) {
+                rawProperties.load(propertiesStream);
+            }
+        }
+        Enumeration<?> e = rawProperties.propertyNames();
```
gituser173: maybe use the entrySet() here to get both name and value

cloudslang-cli/src/main/java/io/cloudslang/lang/cli/SlangBootstrap.java
```diff
@@ -21,9 +28,50 @@
...
+            System.setProperty(key, value);
+        }
+    }
+
+    private static String substitutePropertyReferences(String value) {
+        Matcher mather = SUBSTITUTION_PATTERN.matcher(value);
+        Set<String> variableNames = new HashSet<>();
+        while (mather.find()) {
+            variableNames.add(mather.group(1));
+        }
```
gituser173: maybe split this method into two methods at this line, the above part "findSystemPropertiesNames", the below part "replaceSystemPropertiesNamesWithValues" or something similar. WDYT?

cloudslang-cli/src/main/java/io/cloudslang/lang/cli/SlangBootstrap.java
```diff
@@ -21,9 +28,50 @@
  */
 public class SlangBootstrap {
 
+    private static final String USER_CONFIG_DIR = "configuration";
+    private static final String USER_CONFIG_FILENAME = "cslang.properties";
+    private static final String USER_CONFIG_FILEPATH = USER_CONFIG_DIR + File.separator + USER_CONFIG_FILENAME;
+    private static final String SUBSTITUTION_REGEX = "\\$\\{([^${}]+)\\}"; // ${VAR}
```
gituser173: // ${SystemPropertyName} instead of ${VAR} would be more concise maybe?

Bonczidai:
please note note to delete feature branches - like this one -  in this sprint, they will also need to be merged in v1.0

PR #582, opened 2016-02-29
add default encoding

### LINKED ISSUE #562: Need to know if the input names and values support different language characters
Need to know if when running on a Japanese, German or some other localized environment, the user is able to use the localized characters for input names and input values.
Example of values  噂閏云ボ厭ポ院ｱ, ÄÖÜßÁÀÂÉÊ€äöü°áàâéê¢

gituser173:
Fixes #562


cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/SlangSource.java
```diff
@@ -38,19 +40,33 @@ public static SlangSource fromFile(File file) {
...
         try {
-            source = FileUtils.readFileToString(file, Charset.defaultCharset());
+            source = readFileToString(file);
         } catch (IOException e) {
             throw new RuntimeException("There was a problem reading the yaml file: " + file.getName(), e);
         }
         return new SlangSource(source, file.getName());
     }
 
+    public static String readFileToString(File file) throws IOException {
```
Bonczidai: given the factbthat for now only files that end with `.sl` are considered slang sources (inputs files are not) + this is a general utility method, maybe you should create a new utility class for this method

cloudslang-entities/src/main/java/io/cloudslang/lang/entities/SlangSystemPropertyConstant.java
```diff
@@ -0,0 +1,19 @@
...
+
+/**
+ * User: bancl
+ * Date: 2/29/2016
+ */
+public enum SlangSystemPropertyConstant {
+    CSLANG_ENCODING("cslang.encoding"),
+    LOG4J_CONFIGURATION("log4j.configuration");
+
+    private final String value;
```
Bonczidai: newline here ? :)

cloudslang-tests/src/test/java/io/cloudslang/lang/systemtests/SimpleFlowTest.java
```diff
@@ -133,6 +138,21 @@ public void testFlowWithMissingNavigationFromOperationResult() throws Exception
...
+        URI operations = getClass().getResource("/yaml/print.sl").toURI();
+        String inputValue = FileUtils.readFileToString(new File(getClass().getResource("/inputs/utf8_input.txt").getFile()),
+                StandardCharsets.UTF_8);
+        Map<String, Serializable> inputs = new HashMap<>();
+        inputs.put("input", inputValue);
+
+        SlangSource operationsSource = SlangSource.fromFile(operations);
+        Set<SlangSource> path = Sets.newHashSet(operationsSource);
+        CompilationArtifact compilationArtifact = slang.compile(SlangSource.fromFile(resource), path);
+        trigger(compilationArtifact, inputs, SYS_PROPS);
```
Bonczidai: you could propagate the value through Jython (use publish for task) and verify the flow output

cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/SlangSource.java
Bonczidai: or another approach is to use SlangSource also for loading inputs files (all kinds of files) - we are already doing it in the builder for `.input.yaml` files.. so if you use SlangSource in CompilerHelperImpl.loadMapsFromFiles method we will all be aligned to SlangSource and it will be easier to update in the future -- inthat case the moethods can stay here

PR #583, opened 2016-02-29
Compiler error handling

Bonczidai:


orius123:
Can you please write a short summary for the changes here? 
Why do you need this extra modelling phase and classes?

Bonczidai:
there are cases when the flow is valid Yaml but not necessarily valid from CloudSlang POV - e.g. no workflow section, an invalid input - but we still want to extract data from it, like namespace etc.

so, the solution was to rewrite the precompile stage so it will not fail on the first error but collect all the exceptions - so we will get the partial data + collection of exceptions

Bonczidai:
already in new release branch 0.9.50

PR #595, opened 2016-03-07
Fix error messages

### LINKED ISSUE #594: Need to align preCompileSource() error list details
Currently, the exceptions returned by preCompileSource() method have different formats (some contain the flow name, some not, etc):
- Error compiling testFlow. Flow: testFlow has no workflow property
- Flow must have tasks in its workflow

Need to align the format of the messages, so the list of exception can be displayed in a nice manner.

Bonczidai:
fixes #594 

### DIFF
--- a/cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/modeller/DependenciesHelper.java
+++ b/cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/modeller/DependenciesHelper.java
@@ -120,7 +120,8 @@ public class DependenciesHelper {
         result.addAll(
                 getSystemPropertiesFromPostTaskActionData(
                         task.getPostTaskActionData(),
-                        relevantTransformers
+                        relevantTransformers,
+                        task.getName()
                 )
         );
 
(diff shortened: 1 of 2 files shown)


Bonczidai:
Worked before:

Flow:
```console
java.lang.RuntimeException: java.lang.RuntimeException: For flow 'simple_loop' syntax is illegal.
Under property: 'inputs' there should be a list of values, but instead there is a map.
By the Yaml spec lists properties are marked with a '- ' (dash followed by a space)
```

Op:
```console
java.lang.RuntimeException: java.lang.RuntimeException: For operation 'print' syntax is illegal.
Under property: 'inputs' there should be a list of values, but instead there is a map.
By the Yaml spec lists properties are marked with a '- ' (dash followed by a space)
```

Task:
```console
java.lang.RuntimeException: java.lang.RuntimeException: For task: print_values syntax is illegal.
Task has to many keys under the 'do' keyword,
May happen due to wrong indentation
```

action: n/a - they don't have a name associated

Fixed:
 - your case:
```console
java.lang.RuntimeException: java.lang.RuntimeException: Error compiling simple_loop. Flow: simple_loop has no workflow data
```

@Ligia-Andreica WDYT?

Bonczidai:
now I fixed the problematic case - in the future, if a more comprehensive error message strategy will be required, we'll tackle it then

PR #597, opened 2016-03-09
inspect adds extra quote characters to doc section

### LINKED ISSUE #565: inspect adds extra quote characters to doc section
When running an `inspect` on a file that has quote characters (single or double) the displayed documentation has extra quotes in it. 

For example, see \content\io\cloudslang\base\json\add_value.sl

```
#!!
#! @description: Adds or replaces a value to the given JSON at the keys or indices represented by the json_path.
#!               If the last key in the path does not exist, the key is added as well.
#! @input json_input: JSON data input - Example: '{"k1": {"k2": ["v1", "v2"]}}'
#! @input json_path: path at which to add value represented as a list of keys and/or indices - Example: ["k1","k2",1]
#! @input value: value to associate with key - Example: "v3"
#! @output json_output: JSON with key:value added
#! @output return_result: parsing was successful or not
#! @output return_code: "0" if parsing was successful, "-1" otherwise
#! @output error_message: error message if there was an error when executing, empty otherwise
#! @result SUCCESS: parsing was successful (return_code == '0') and value was added
#! @result FAILURE: parsing was unsuccessful or the path does not exist
#!!#
```

image

gituser173:
Fixes #565 

### DIFF
--- a/cloudslang-cli/src/main/java/io/cloudslang/lang/cli/utils/MetadataHelperImpl.java
+++ b/cloudslang-cli/src/main/java/io/cloudslang/lang/cli/utils/MetadataHelperImpl.java
@@ -12,19 +12,10 @@ import io.cloudslang.lang.api.Slang;
 import io.cloudslang.lang.compiler.SlangSource;
 import io.cloudslang.lang.compiler.modeller.model.Metadata;
 import org.apache.commons.lang.Validate;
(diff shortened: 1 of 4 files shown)


Bonczidai:
:+1: looks good - please handle @SamMarkowitz 's comment about empty tags and we're good to go

PR #609, opened 2016-03-16
Context visibility

### LINKED ISSUE #568: Remove `self` keyword - context visibility
Consider case of a task-subflow (it also applies to operation-action, aggregate section etc.) in the publish section.

Behavior now:
- visible context: subflow context
- reference flow contex: with `self` keyword

Usage is rare and confusing, so we suggest to go with:

Behavior after the change:
- visible context: flow context + subflow outputs (takes priority)
- we will fire a warning when var name is identical in subflow outputs and flow context
  - checking the conflicts at compile time may be an option - but we need to take into account flow inputs and publish sections (all the possibilities that modify flow context - not sure it's a good idea)

@CloudSlang/slang-functional-reviewers please give feedback

Bonczidai:
fixes #568 

### DIFF
--- a/cloudslang-runtime/src/main/java/io/cloudslang/lang/runtime/bindings/OutputsBinding.java
+++ b/cloudslang-runtime/src/main/java/io/cloudslang/lang/runtime/bindings/OutputsBinding.java
@@ -14,8 +14,8 @@ import io.cloudslang.lang.entities.SystemProperty;
 import io.cloudslang.lang.entities.bindings.Output;
 
 import io.cloudslang.lang.entities.utils.ExpressionUtils;
+import io.cloudslang.lang.entities.utils.MapUtils;
 import io.cloudslang.lang.runtime.bindings.scripts.ScriptEvaluator;
-import org.apache.commons.collections4.MapUtils;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Component;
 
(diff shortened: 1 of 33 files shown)


Bonczidai:
since post-task bindings can use task arguments we need to:
- save them before subflow invocation
- retrieve them after subflow invocation

for that I chose to use the [langVariables](https://github.com/CloudSlang/cloud-slang/blob/b4e9a9e86a1d9e77ec9b71eb79f26bc702851c6c/cloudslang-runtime/src/main/java/io/cloudslang/lang/runtime/env/Context.java#L23) container (where we also store the loop condition)

WDYT? @oritstone @orius123

orius123:
Sounds good to me.
where was it stored till now?
how does the user access them?

Bonczidai:
the result of task argument binding is passed to the subflow in `RunEnv`'s `callArguments` field, now we need them on the same level (Context is saved in `ContextStack`)

Bonczidai:
wait with the merge for now

cloudslang-runtime/src/main/java/io/cloudslang/lang/runtime/bindings/OutputsBinding.java
```diff
@@ -79,4 +69,18 @@
         return outputs;
     }
 
+    private Map<String, Serializable> mergeContexts(
```
gituser173: this helper method seems to be duplicated in ResultsBinding

gituser173:
@Bonczidai please take care of the comment, otherwise it looks good :+1:

PR #623, opened 2016-03-29
Function enhancements

Bonczidai:
- support get without default param
- skip whitespaces in expressions


Bonczidai:
TODO: consider trimming the expr after loading

Bonczidai:
CloudSlang expressions with whitespaces like `${ <spaces> expr <spaces> }` preserve the chars, so the actual expression will contain space characters, so we'll get `eval( <spaces> expr <spaces> )`

i suggest to skip the whitespace chars, so we'll get `${ <spaces> expr <spaces> }` -> `eval(expr)`

do you have anything against? @meshipeer @gituser173 @oritstone

gituser173:
Looks good :+1: 
can be merged after the build passes

PR #650, opened 2016-04-14
Added required, Overidable, no default value validation

sekler:
Added required, Overidable, no default value validation 
by @genadi-hp and @sekler 

sign-off @Bonczidai


Bonczidai:
:+1: thanks guys for the contribution!

my feedback:
- need to remove Java 8 features - we are using Java 7
- extracting the inputs from all the classpath files + matching only to current flow (no recursivity) -> more efficient would be to iterate based on current flow steps
- IMHO we should validate recusively for each executable that is reachable from current flow (subset of all the `"path"` supplied)

I did this changes in a wrapper branch (your commits appear there) - you can track the progress in #655

PR #651, opened 2016-04-14
Fail compilation if step is not reachable

### LINKED ISSUE #305: fail compilation if task is not wired
if we have a flow
```yaml
task1:
  ...
  navigate:
    SUCCESS: SUCCESS
    FAILURE: FAILURE
task2:
  ...
```
because task2 is not wired and therefore not accessible at runtime we don't compile it..

ifatgv4:
fixes #305
I added to ExecutableBuilder.compileWorkFlow() code that finds all the tasks that are not reachable. (This code is similar to existing code in ExecutionPlanBuilder, that compiles the reachable steps).

I also added 4 tests for 4 new .sl files:

1. unreachable_tasks.sl - contains two unreachable tasks. The preCompilation returns two unreachable task exceptions
2. unreachable_on_failure_task - all tasks are reachable, except for the one in onFailure - compilation succeeds
3. unreachable_task_reachable_from_on_failure - one task is unreachable from other tasks, but is reachable from the onFailure task - compilation succeeds
4. unreachable_tasks_one_reachable_from_on_failure - contains two tasks which are unreachable from other tasks, but one of them is reachable from the onFailure task - The compilation returns only one unreachable task exception, for the one that is not reachable

### DIFF
--- a/cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/modeller/ExecutableBuilder.java
+++ b/cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/modeller/ExecutableBuilder.java
@@ -8,6 +8,7 @@
  */
 package io.cloudslang.lang.compiler.modeller;
 
+import ch.lambdaj.Lambda;
 import io.cloudslang.lang.compiler.SlangTextualKeys;
 import io.cloudslang.lang.compiler.modeller.model.Action;
 import io.cloudslang.lang.compiler.modeller.model.Flow;
@@ -42,6 +43,7 @@ import java.util.Deque;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Iterator;
+import java.util.LinkedHashMap;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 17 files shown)


cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/modeller/ExecutableBuilder.java
```diff
@@ -480,4 +488,33 @@ private static String resolveReferenceID(String rawReferenceID, Map<String, Stri
...
+                    unreachableSteps.add(step);
+                }
+            }
+        }
+        return unreachableSteps;
+    }
+
+    private void getReachableSteps(Step step, Deque<Step> steps, Map<String, Step> reachableSteps) {
+        reachableSteps.put(step.getName(), step);
+        for (Map<String, String> map : step.getNavigationStrings()) {
```
Bonczidai: please update from master - method signature changed: `public List<Map<String, String>> getNavigationStrings() {
        return navigationStrings;
    }`

Bonczidai:
@ifatgv4 looks good :+1: please handle my small comment

Bonczidai:
:ok_hand:

PR #655, opened 2016-04-15
Inputs validation

### LINKED ISSUE #625: Validate task data against dependency data
to discuss  which validations we would like to apply:
- `overridable, required, without default value` subflow inputs are supplied

Bonczidai:
fixes #625 
closes #650 


ifatgv4:
I reviewed the change and it seems OK except for one thing - the validation is done during compile and I think it should be done during pre-compile

Bonczidai:
thanks @ifatgv4 -- I don't think we can do it in `preCompile` since preCompile deals with one executable -- in this validation we need all the dependencies which are only available after preCompiling the sources

PR #661, opened 2016-04-19
use latest score version and also use new artifacts for spring-io pro…

i022098:
you need first to take my PR from score and than use this PR that take score latest version


Bonczidai:
migrated to #663

PR #666, opened 2016-04-21
System Properties Should contain Properties as List

### LINKED ISSUE #644: System Properties Should contain Properties as List
System Properties now contains map of properties.
We need the compiler to fail upon two properties with the same key.
Therefore, the properties entry should be a list instead of a map

Bonczidai:
fixes #644 


gituser173:
@Bonczidai you can merge :+1:

PR #668, opened 2016-04-26
Added support to CloudSlang for dependencies on action level

### LINKED ISSUE #630: Support backward compatibility - versioning
- backward compatibility *may* be affected by the cloudslang versioning plans (e.g. adding version tag in DSL)
- visual data in flows  - [SOLVED](https://github.com/CloudSlang/cloud-slang/issues/643)

genadi-hp:
fixes #630 

Nothing to do with them until we have dependencies resolution. Dependencies for java are still optional for the same reason.

Signed-off-by: Genadi Rabinovich <genadi@hpe.com>


cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/modeller/transformers/DependenciesTransformer.java
```diff
@@ -0,0 +1,37 @@
...
+    @Override
+    public List<String> transform(List<String> rawData) {
+        return (rawData == null) ? Collections.<String>emptyList() : rawData;
+    }
+
+    @Override
+    public List<Scope> getScopes() {return Arrays.asList(Scope.ACTION);}
+
+    @Override
+    public String keyToTransform() {return SlangTextualKeys.ACTION_DEPENDENCIES;}
```
Bonczidai: I think it would be better to call the class `ActionDependenciesTransformer` so the scope would be clear, in that case you could even return null here since the key can be taken from class name

cloudslang-runtime/src/main/java/io/cloudslang/lang/runtime/steps/ActionExecutionData.java
```diff
@@ -55,6 +55,7 @@
     public void doAction(@Param(ScoreLangConstants.RUN_ENV) RunEnvironment runEnv,
                          @Param(ExecutionParametersConsts.NON_SERIALIZABLE_EXECUTION_DATA) Map<String, Object> nonSerializableExecutionData,
                          @Param(ScoreLangConstants.ACTION_TYPE) ActionType actionType,
+                         @Param(ScoreLangConstants.ACTION_DEPENDENCIES) List<String> actionDependencies,
```
Bonczidai: best practice: for an ExecutionStep parameter the name (`actionDependencies`) should be the same as the `@Param` value (dependencies), however `@Param` counts

I suggest either `actionDependencies` -> `dependencies` or change `ScoreLangConstants.ACTION_DEPENDENCIES` -> `actionDependencies`

cloudslang-compiler/src/test/java/io/cloudslang/lang/compiler/modeller/transformers/DependenciesTransformerTest.java
```diff
@@ -0,0 +1,38 @@
...
+
+    @Test
+    public void testTransformWithEmptyDependencies() throws Exception {
+        DependenciesTransformer transformer = new DependenciesTransformer();
+
+        List<String> list = Lists.newArrayList();
+        List<String> transform = transformer.transform(list);
+        assertTrue(transform.isEmpty());
+        assertEquals(list, transform);
+    }
```
Bonczidai: maybe add a test case for transforming non-empty list

Bonczidai:
:+1: very neat @genadi-hp 

added some minor comments, please take a look

cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/modeller/transformers/DependenciesTransformer.java
genadi-hp: This could be a good idea, but there were talks about possibility to have dependencies on upper (operation) level as well. We may want to have multiple scope for dependencies and in this case DependenciesTransformer may be more generic name.

cloudslang-runtime/src/main/java/io/cloudslang/lang/runtime/steps/ActionExecutionData.java
genadi-hp: Ok, thanks, will rename actionDependencies -> dependencies

cloudslang-compiler/src/test/java/io/cloudslang/lang/compiler/modeller/transformers/DependenciesTransformerTest.java
genadi-hp: Thanks, will be added

genadi-hp:
These changes are no more relevant for us, I am closing this PR

PR #672, opened 2016-05-03
added score-xml to poms

SamMarkowitz:


SamMarkowitz:
@Bonczidai I'm not sure why this build is failing. It doesn't seem related to me. Ideas?

Bonczidai:
we have a sporadic failure: #365 which is getting nastier as time it grows up - we need to tackle it soon ( solution is t restart build :( )

PR #674, opened 2016-05-03
Add validate model with dependencies

oritstone:
Add a new method to the Slang compiler API that allows to validate an Executable model against the models of it dependencies
For now we validate:
- That all the results of the dependency have matching navigations in the step
- That all required inputs of the dependency with no default value, have matching inputs in the task

### DIFF
--- a/cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/SlangCompiler.java
+++ b/cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/SlangCompiler.java
@@ -48,8 +48,8 @@ public interface SlangCompiler {
      * Validate that the given {@Link io.cloudslang.lang.compiler.modeller.model.Executable} is valid regarding
      * its wiring to its dependencies
      * Current validations:
-     *      - Validates that required inputs of the dependecy have a matching input in the task
-     *      - Validate that every result of teh dependency has a matching navigation in the task
+     *      - Validates that required inputs of the dependency have a matching input in the step
+     *      - Validate that every result of teh dependency has a matching navigation in the step
      * @param slangModel the CloudSlang model to validate
      * @param dependenciesModels the CloudSlang models of the direct dependencies
      * @return a list of the exceptions that were found (if any)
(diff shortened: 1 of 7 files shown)


cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/SlangCompiler.java
```diff
@@ -43,6 +44,18 @@
      */
     ExecutableModellingResult preCompileSource(SlangSource source);
 
+    /**
+     * Validate that the given {@Link io.cloudslang.lang.compiler.modeller.model.Executable} is valid regarding
+     * its wiring to its dependencies
+     * Current validations:
+     *      - Validates that required inputs of the dependecy have a matching input in the task
```
Bonczidai: step :) not task

cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/SlangCompiler.java
```diff
@@ -43,6 +44,18 @@
...
+     *      - Validate that every result of teh dependency has a matching navigation in the task
```
Bonczidai: step :) not task

cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/scorecompiler/ScoreCompiler.java
```diff
@@ -35,4 +36,16 @@
      */
     CompilationArtifact compile(Executable source, Set<Executable> path);
 
+    /**
+     * Validate that the given {@Link io.cloudslang.lang.compiler.modeller.model.Executable} is valid regarding
+     * its wiring to its dependencies
+     * Current validations:
+     *      - Validates that required inputs of the dependecy have a matching input in the task
```
Bonczidai: please propagate the comments here too

Bonczidai:
:ok_hand: 

added some minor comments just for the sake of not being fast-forward :smiling_imp: -- please take a look

PR #677, opened 2016-05-06
Change op hierarchy

### LINKED ISSUE #675: BC - change operation hierarchy
```
•	The action hierarchy will be changed. The ‘action’ key will be removed and instead there will be:
o	java_action
o	phyton_action
•	As a result of the difference in the action key, we are not oblige to stick to the same keys underneath.
•	Java action hierarchy will look as follow:
Java_action
                gav: ‘g:a:v’
                class_name:
                method_name:
•	Phyton action hierarchy will look as follow (dependencies tag is not mandatory):
phyton_action
                dependencies:
                                -‘g1:a1:v1’
                                -‘g2:a2:v2’
                script:
•	The template will be written as described above.
```

Bonczidai:
fixes #675


Bonczidai:
first commit: production changes (https://github.com/CloudSlang/cloud-slang/pull/677/commits/840ba86857cf9f6fdcbdc0f98f2078211d2d61d5)
second commit: test changes (https://github.com/CloudSlang/cloud-slang/pull/677/commits/19ba6a84484b4c574f3ae75a9f942bafa18e16dd)

cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/SlangTextualKeys.java
```diff
@@ -29,10 +29,13 @@
...
 
     //action
-    String JAVA_ACTION = "java_action";
-
-    //operation
-    String ACTION_KEY = "action";
+    String JAVA_ACTION_KEY = "java_action";
+    String JAVA_ACTION_CLASS_NAME_KEY = "class_name";
+    String JAVA_ACTION_METHOD_NAME_KEY = "method_name";
+    String JAVA_ACTION_GAV_KEY = "gav";
```
oritstone: I do not think that it should be called gav... It's a Maven term, not a Java term
I think we should call it something like `artifact:` and maybe support a value in the syntax of gav, or the simple artifact name in case the CloudSlang user does not use Maven

cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/modeller/ExecutableBuilder.java
```diff
@@ -83,10 +82,15 @@
 
     private List<Transformer> preExecTransformers;
     private List<Transformer> postExecTransformers;
-    private List<String> execAdditionalKeywords = Arrays.asList(SlangTextualKeys.ACTION_KEY, SlangTextualKeys.WORKFLOW_KEY, SlangTextualKeys.EXECUTABLE_NAME_KEY);
+    private List<String> execAdditionalKeywords = Arrays.asList(
+            SlangTextualKeys.JAVA_ACTION_KEY,
+            SlangTextualKeys.PYTHON_ACTION_KEY,
+            SlangTextualKeys.WORKFLOW_KEY,
+            SlangTextualKeys.EXECUTABLE_NAME_KEY
```
oritstone: Don't we want to add also the additional keywords of class_name, method_name, script, etc... ?

cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/modeller/transformers/AbstractTransformer.java
```diff
@@ -0,0 +1,54 @@
...
+public abstract class AbstractTransformer {
+
+    public static final String MISSING_KEYS_ERROR_MESSAGE_PREFIX = "Following keys are missing: ";
+    public static final String INVALID_KEYS_ERROR_MESSAGE_PREFIX = "Following keys are invalid: ";
+
+    protected void validateKeySet(
+            Set<String> keySet,
+            Set<String> mandatoryKeys,
+            Set<String> optionalKeys) {
+        Validate.notNull(keySet);
```
oritstone: I do not think we should validate and throw exception if the sets are empty. 
It is not an error if they are, it's only that we cannot validate anything

cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/SlangTextualKeys.java
oritstone: I take it back :)
I understand we are going to use Maven inside CloudSlang as the optional mechanism for retrieving artifacts. In that case, `gav` is the correct term

cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/scorecompiler/ExecutionStepFactory.java
```diff
@@ -91,17 +91,19 @@ public ExecutionStep createActionStep(Long index, Map<String, Serializable> acti
         Map<String, Serializable> actionData = new HashMap<>();
         ActionType actionType;
 
-        @SuppressWarnings("unchecked") Map<String, String> javaActionData = (Map<String, String>) actionRawData.get(SlangTextualKeys.JAVA_ACTION);
-        @SuppressWarnings("unchecked") String pythonScript = (String) actionRawData.get(ScoreLangConstants.PYTHON_SCRIPT_KEY);
+        @SuppressWarnings("unchecked")
+        Map<String, Serializable> javaActionData = (Map<String, Serializable>) actionRawData.get(SlangTextualKeys.JAVA_ACTION_KEY);
```
oritstone: Why isn't it still `Map<String, String>`?

cloudslang-compiler/src/test/java/io/cloudslang/lang/compiler/CompileOperationTest.java
```diff
@@ -90,15 +90,15 @@ public void testPreCompileOperationBasic() throws Exception {
     public void testCompileOperationMissingClassName() throws Exception {
         URL resource = getClass().getResource("/corrupted/operation_missing_class_name.sl");
         exception.expect(RuntimeException.class);
-        exception.expectMessage("className");
+        exception.expectMessage("class_name");
```
oritstone: Why not to use the constant you added?
Also for `method_name` below

cloudslang-compiler/src/test/java/io/cloudslang/lang/compiler/modeller/ExecutableBuilderTest.java
```diff
@@ -312,43 +312,6 @@ public void operationWithEmptyActionDataThrowException() throws Exception {
         transformToExecutable(mockParsedSlang, operationName, executableRawData);
     }
```
oritstone: Don't we want some tests for invalid keys in action?

oritstone:
Good work Levi!
Wrote some small comments

cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/scorecompiler/ExecutionStepFactory.java
Bonczidai: right

cloudslang-compiler/src/test/java/io/cloudslang/lang/compiler/CompileOperationTest.java
Bonczidai: right

cloudslang-compiler/src/test/java/io/cloudslang/lang/compiler/modeller/ExecutableBuilderTest.java
Bonczidai: because of the restructuring, now the logic of key validation is moved to the java/python transformers

tests are added there: e.g. https://github.com/CloudSlang/cloud-slang/pull/677/files#diff-6e4240d6335bba4b8445785e3e39aa6dR87

cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/modeller/transformers/AbstractTransformer.java
Bonczidai: agree, for empty sets it works - I am validating here `null` refs so I will not get `NullPointerExceptions`

if you think it should be changed, what would be the suggesiton?

Bonczidai:
thanks for the review - addressed the issues - please take a look

cloudslang-compiler/src/test/java/io/cloudslang/lang/compiler/modeller/ExecutableBuilderTest.java
oritstone: :+1:

cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/modeller/transformers/AbstractTransformer.java
oritstone: So it's OK

cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/scorecompiler/ExecutionStepFactory.java
```diff
@@ -91,17 +91,19 @@ public ExecutionStep createActionStep(Long index, Map<String, Serializable> acti
         Map<String, Serializable> actionData = new HashMap<>();
         ActionType actionType;
 
-        @SuppressWarnings("unchecked") Map<String, String> javaActionData = (Map<String, String>) actionRawData.get(SlangTextualKeys.JAVA_ACTION);
-        @SuppressWarnings("unchecked") String pythonScript = (String) actionRawData.get(ScoreLangConstants.PYTHON_SCRIPT_KEY);
+        @SuppressWarnings("unchecked")
+        Map<String, String> javaActionData = (Map<String, String>) actionRawData.get(SlangTextualKeys.JAVA_ACTION_KEY);
+        @SuppressWarnings("unchecked")
+        Map<String, Serializable> pythonActionData = (Map<String, Serializable>) actionRawData.get(SlangTextualKeys.PYTHON_ACTION_KEY);
```
oritstone: I think that also `pythonActionData` should be `Map<String, String>`

oritstone:
Another small comment and that's it :)

cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/scorecompiler/ExecutionStepFactory.java
Bonczidai: python action data contains the script (string) but also dependencies (which will be list -- multiple py dependencies)

Bonczidai:
thanks! replied to your comment - now I dare to merge it :)

PR #687, opened 2016-05-18
Support sensitive data

### LINKED ISSUE #506: encrypted inputs \ outputs

ifatgv4:
fixes #506
Add support for "sensitive" in Inputs and "transitive sensitivity" in Python script Outputs.


cloudslang-entities/src/main/java/io/cloudslang/lang/entities/bindings/values/PyObjectValueProxyFactory.java
```diff
@@ -0,0 +1,120 @@
...
+            this.value = ValueFactory.create(content, sensitive);
+            this.pyObject = pyObject;
+            this.accessed = false;
+        }
+
+        @Override
+        public Object invoke(Object self, Method thisMethod, Method proceed, Object[] args) throws Throwable {
+            if (thisMethod.getName().equals("isAccessed")) {
+                return accessed;
+            } else if (thisMethod.getDeclaringClass().isAssignableFrom(Value.class)) {
```
Bonczidai: shouldn't this be vice-versa? `value assignable from class`

Bonczidai:
nice direction :+1:

cloudslang-entities/src/main/java/io/cloudslang/lang/entities/bindings/values/PyObjectValueProxyFactory.java
ifatgv4: Yes, you are right :)

I fixed it.

ifatgv4:
Thanks, but I think there is a better direction than this one. At first I thought we needed transitivity for Script execution as well, so I didn't even go there, but now that I understand that we only need transitivity for expression evaluation I think it would be better to write a parser. It is really not difficult.

I will send you the mail I sent Avigail and Meshi about it.

cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/SlangTextualKeys.java
```diff
@@ -46,9 +46,10 @@
     String PARALLEL_LOOP_KEY = "parallel_loop";
 
     //inputs
+    String VALUE_KEY = "value";
```
Bonczidai: maybe this tag should be aligned with input's `default` keyword - WDYT?

Bonczidai:
:) My couple of cents about the tests:
- maybe break the huge test cases into several ones, will ease the debugging in case of failures
- you can consider also the following test cases:
  - for loop
  - parallel loop
  - op with java action
  - sensitive values in op result conditions
  - test the SD does not appear in events / logs (we can also consider adding tests here to: https://github.com/CloudSlang/test-functional -- if you think this is the case, please open an issue there)

cloudslang-tests/src/test/java/io/cloudslang/lang/systemtests/ParallelLoopFlowsTest.java
```diff
@@ -49,6 +55,30 @@ public void testFlowWithParallelLoop() throws Exception {
...
     @Test
+    public void testFlowWithSensitiveInputParallelLoop() throws Exception {
+        URI resource = getClass().getResource("/yaml/loops/parallel_loop/sensitive_input_parallel_loop.sl").toURI();
+        URI operation1 = getClass().getResource("/yaml/loops/parallel_loop/print_branch.sl").toURI();
+        Set<SlangSource> path = Sets.newHashSet(SlangSource.fromFile(operation1));
+
+        RuntimeInformation runtimeInformation = triggerWithData(SlangSource.fromFile(resource), path);
+
+        List<StepData> branchesData = extractParallelLoopData(runtimeInformation);
+        Assert.assertEquals("incorrect number of branches", 3, branchesData.size());
```
Bonczidai: isn't these tests suppose to validate the some data  - like the flow inputs/outputs are sensitive?

cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/SlangTextualKeys.java
ifatgv4: I did want to use the same name, but it didn't seem right.

default is wrong for outputs, because it's not the default value, it's the actual calculated value
value is wrong for inputs, because it's not the value, just the default in case a real one was not provided

Unless you have a good suggestion for both, I guess they will have to be different

Bonczidai:
:fireworks: :fireworks: :fireworks:

cloudslang-tests/src/test/java/io/cloudslang/lang/systemtests/ParallelLoopFlowsTest.java
ifatgv4: You are right. I will add that

PR #689, opened 2016-05-19
compiler error improvement

### LINKED ISSUE #685: compiler error improvment
regarding unsupported tags in the DSL

gituser173:
Fixes #685


Bonczidai:
- found one case: under python_action tag (I think we should use the same error message there)
- please make sure to test each case (for each level)
- shouldn't we add the link in the error message? - e.g. you could read from a property file the changelog for now

:+1:

gituser173:
it was decided that we should not add a link in the error message

Bonczidai:
please also merge from master (sensitive + versioning)

PR #694, opened 2016-05-25
Inputs initialized with null before binding

### LINKED ISSUE #620: Inputs initialized with null before binding
the inputs / arguments binding initializes the input name with null before binding:

after ` - input1: ${ input1 }` will have null value after binding instead of an exception
IMHO this case is no longer valid and it should throw an exception.
For this case we have `- input1` syntax and the `get` function

gituser173:
Fixes  #620


cloudslang-entities/src/main/java/io/cloudslang/lang/entities/bindings/Input.java
```diff
@@ -110,10 +117,16 @@ public InputBuilder(String name, Serializable value) {
 			encrypted = false;
 			required = true;
 			privateInput = false;
+            defaultSpecified = false;
```
Bonczidai: pay attention to indent :)

Bonczidai:
:+1: looks cool - please also add a system test

```
    - null_input:
        default: ${null_input}
        required: false
```

which is supposed to fail

Bonczidai:
please also merge from master (sensitive + versioning)

cloudslang-entities/src/main/java/io/cloudslang/lang/entities/bindings/Input.java
```diff
@@ -29,6 +29,7 @@
 
 	private boolean required;
 	private boolean privateInput;
+    private boolean defaultSpecified;
```
Bonczidai: do you still need this property after the changes?

Bonczidai:
please only assign once build passes : )

Bonczidai:
:+1: we'll merge it after sensitive data + versioning fixes are in

PR #698, opened 2016-05-26
Fix bug in direct dependencies validation:

oritstone:
The validation should be done only on direct dependency, not recursively. It's a shallow validation


Bonczidai:
:ok_hand:

PR #728, opened 2016-06-08
Builder action args receive input objects

### LINKED ISSUE #727: Builder: action args receive Input objects
instead of String for example

gituser173:
Fixes #727


Bonczidai:
thanks @gituser173

PR #729, opened 2016-06-08
Allow navigation to on_failure even if on_failure step not exists

### LINKED ISSUE #716: Allow navigation to on_failure even if on_failure step not exists
In order for the WS to be able to recognize cases were the step navigation points to FAILURE, but actually to present it in the WS as it actually points to the on-failure we need to be able to set navigation to on-failure even if the on_failure step not exists.
the behavior in this case should be like it navigate to FAILURE result of the flow

gituser173:
Fixes #716

### DIFF
--- a/cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/modeller/ExecutableBuilder.java
+++ b/cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/modeller/ExecutableBuilder.java
@@ -492,11 +492,7 @@ public class ExecutableBuilder {
                 Map<String, String> transformedNavigation = new HashMap<>(navigation);
                 if (navigationEntry.getValue().equals(ON_FAILURE_KEY)) {
                     if (StringUtils.isEmpty(onFailureStepName)) {
-                        throw new RuntimeException(
-                                "Failed to compile step: '" + stepName +
-                                        "'. Navigation: '" + navigationEntry.getKey() + " -> " + navigationEntry.getValue()
-                                        + "' is illegal. 'on_failure' section is not defined."
-                        );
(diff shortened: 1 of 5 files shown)


cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/modeller/ExecutableBuilder.java
```diff
@@ -492,11 +492,7 @@ private void replaceOnFailureReference(
                 Map<String, String> transformedNavigation = new HashMap<>(navigation);
                 if (navigationEntry.getValue().equals(ON_FAILURE_KEY)) {
                     if (StringUtils.isEmpty(onFailureStepName)) {
-                        throw new RuntimeException(
-                                "Failed to compile step: '" + stepName +
```
Bonczidai: step_name becomes unused right?

Bonczidai:
do we have a system test for this case?

cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/modeller/ExecutableBuilder.java
```diff
@@ -492,11 +492,7 @@ private void replaceOnFailureReference(
...
-                                        "'. Navigation: '" + navigationEntry.getKey() + " -> " + navigationEntry.getValue()
-                                        + "' is illegal. 'on_failure' section is not defined."
-                        );
+                        transformedNavigation.put(navigationEntry.getKey(), ON_FAILURE_KEY);
```
Bonczidai: can we put `FAILURE` here?

Bonczidai: maybe that way you don;t need to even modify EPB

PR #730, opened 2016-06-13
on_failure section should end with FAILURE

### LINKED ISSUE #722: on_failure section should end with FAILURE
on_failure section should not allow any navigation and end the flow with FAILURE

gituser173:
Fixes #722 


cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/modeller/ExecutableBuilder.java
```diff
@@ -307,6 +309,31 @@ private Workflow getOnFailureWorkflow(List<Map<String, Map<String, Object>>> wor
...
+            errors.add(new RuntimeException("Flow: '" + execName + "' syntax is illegal.\nThe step below 'on_failure' property should not contain a \"navigate\" section."));
+        }
+        List<Map<String, String>> failureNavigationSection = getFailureNavigationSection();
+        onFailureStep.getValue().put(NAVIGATION_KEY, failureNavigationSection);
+    }
+
+    private List<Map<String, String>> getFailureNavigationSection() {
+        List<Map<String, String>> failureNavigationSection = new ArrayList<>();
+        Map<String, String> success = new HashMap<>();
+        success.put(ScoreLangConstants.SUCCESS_RESULT, ScoreLangConstants.SUCCESS_RESULT);
```
Bonczidai: does this work? should end with FAILURE in any case, should take into account custom results, right?

Bonczidai:
do we have explicit test cases for:
- compilation fails when navigate is found
- flow ends with failure when subflow has default / subflow has explicit results

?

cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/modeller/ExecutableBuilder.java
```diff
@@ -307,6 +309,31 @@ private Workflow getOnFailureWorkflow(List<Map<String, Map<String, Object>>> wor
...
+        List<Map<String, String>> failureNavigationSection = getOnFailureNavigationSection();
+        onFailureStep.getValue().put(NAVIGATION_KEY, failureNavigationSection);
+    }
+
+    private List<Map<String, String>> getOnFailureNavigationSection() {
+        List<Map<String, String>> failureNavigationSection = new ArrayList<>();
+        Map<String, String> success = new HashMap<>();
+        success.put(ScoreLangConstants.SUCCESS_RESULT, ScoreLangConstants.FAILURE_RESULT);
+        Map<String, String> failure = new HashMap<>();
+        failure.put(ScoreLangConstants.FAILURE_RESULT, ScoreLangConstants.FAILURE_RESULT);
```
Bonczidai: I think you should not put any navigation here, since you will set the actual results later on in socrecompiler (and the subflow may not have these results)

cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/scorecompiler/ScoreCompilerImpl.java
```diff
@@ -65,6 +69,8 @@ public CompilationArtifact compile(Executable executable, Set<Executable> path)
             //than we match the references to the actual dependencies
             filteredDependencies = dependenciesHelper.matchReferences(executable, availableExecutables);
 
+            handleOnFailureCustomResults(availableExecutables);
```
Bonczidai: I think we can use here filteredDependencies and that way we don't have to recalculate it in `handleOnFailureCustomResults` each time - WDYT?

Bonczidai:
:+1: please handle comments

cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/scorecompiler/ScoreCompilerImpl.java
gituser173: no, because we filter dependencies based on a different executable each time

Bonczidai:
:ok_hand:

PR #741, opened 2016-06-15
Webstudio - mark flow without any results

### LINKED ISSUE #724: Webstudio - mark flow without any results
Hello, 

I'm from OO Guy team, 
We have issue while we trying to create flow without results, 
The issue is - When we create new flow (flow1) and adding one or more steps (operation or flow) and save it without dragging results, then we create another flow (flow2) and dragging the first flow (flow1) to it,
the flow as step created with default ports, Failure and Success that actually these are the default results , although there is no results for flow1.  
We are expected that if there are no results (yaml contains section with 'results: []') the server will return an empty array of results instead of the defaults.

Please let me know if you need further information regarding this issue
Thanks,
Haim

gituser173:
Fixes #724


Bonczidai:
- can you please add a test for flow case
- IMHO this case (`results: []`) shouldn't even pass pre-compile phase..

PR #753, opened 2016-06-26
Change the file encoding used for reading Slang sources, from the JRE…

hprotem:
… default (which in many cases is a regional charset, 1990s-style) to the modern UTF-8. Note: As before, this can be overridden with the "cslang.encoding" system property (though it is probably a better idea to just use UTF-8 in all sources).

### DIFF
--- a/cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/SlangSource.java
+++ b/cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/SlangSource.java
@@ -21,6 +21,7 @@ import java.io.File;
 import java.io.IOException;
 import java.net.URI;
 import java.nio.charset.Charset;
+import java.nio.charset.StandardCharsets;
 
 public class SlangSource {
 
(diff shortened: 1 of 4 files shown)


cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/SlangSource.java
```diff
@@ -55,12 +55,11 @@ private static String readFileToString(File file) throws IOException {
     }
 
     private static Charset getCharset() {
-        Charset charset = Charset.defaultCharset();
         String cslangEncoding = System.getProperty(SlangSystemPropertyConstant.CSLANG_ENCODING.getValue());
-        if (!StringUtils.isEmpty(cslangEncoding)) {
-            charset = Charset.forName(cslangEncoding);
+        if (StringUtils.isEmpty(cslangEncoding)) {
+            cslangEncoding = "UTF-8";
```
Bonczidai: I think we should define a const for `"UTF-8"` - maybe `SlangSystemPropertyConstant` enum is a good place - WDYT?

cloudslang-runtime/src/main/java/io/cloudslang/lang/runtime/configuration/SlangRuntimeSpringConfig.java
```diff
@@ -32,7 +32,9 @@
 
     private static void setPythonIOEncoding() {
         String encodingValue = System.getProperty(SlangSystemPropertyConstant.CSLANG_ENCODING.getValue());
-        if (!StringUtils.isEmpty(encodingValue))
-            System.getProperties().setProperty(PySystemState.PYTHON_IO_ENCODING, encodingValue);
+        if (StringUtils.isEmpty(encodingValue)) {
+            encodingValue = "UTF-8";
```
Bonczidai: and use the const here too

Bonczidai:
:ok_hand:

PR #768, opened 2016-07-06
Case insensitive

### LINKED ISSUE #745: CloudSlang - change to case insenstive
System properties: CloudSlang is case sensitive 
Flows&Operations: it not been checked because the operating system is case insensitive. 

There is a need to change the behavior to be **case insensitive**, that mean that we need to block the option to add 2 system properties with the same name regardless to upper/lower case.
in addition, check the issue regarding flows & operation and block there too.

Bonczidai:
fixes #745 

### DIFF
--- a/cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/SlangCompilerImpl.java
+++ b/cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/SlangCompilerImpl.java
@@ -14,6 +14,7 @@ import io.cloudslang.lang.compiler.modeller.result.ExecutableModellingResult;
 import io.cloudslang.lang.compiler.parser.YamlParser;
 import io.cloudslang.lang.compiler.parser.model.ParsedSlang;
 import io.cloudslang.lang.compiler.scorecompiler.ScoreCompiler;
+import io.cloudslang.lang.compiler.validator.CompileValidator;
 import io.cloudslang.lang.entities.CompilationArtifact;
 import io.cloudslang.lang.entities.SystemProperty;
 import io.cloudslang.lang.entities.utils.SetUtils;
@@ -55,6 +56,9 @@ public class SlangCompilerImpl implements SlangCompiler {
     @Autowired
     private ScoreCompiler scoreCompiler;
 
+    @Autowired
+    private CompileValidator compileValidator;
+
     @Override
     public CompilationArtifact compile(SlangSource source, Set<SlangSource> path) {
 
@@ -62,9 +66,18 @@ public class SlangCompilerImpl implements SlangCompiler {
 
         //we transform also all of the files in the given path to model objects
         Set<Executable> pathExecutables = new HashSet<>();
+
+        Map<SlangSource, Executable> executablePairs = new HashMap<>();
+        executablePairs.put(source, executable);
+
         if (CollectionUtils.isNotEmpty(path)) {
             for (SlangSource pathSource : path) {
-                pathExecutables.add(preCompile(pathSource));
+                Executable preCompiledPathSource = preCompile(pathSource);
+
+                compileValidator.validateNoDuplicateExecutablesBasedOnFQN(preCompiledPathSource, pathSource, executablePairs);
+
+                pathExecutables.add(preCompiledPathSource);
+                executablePairs.put(pathSource, preCompiledPathSource);
             }
         }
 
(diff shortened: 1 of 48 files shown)


Bonczidai:
- validation for duplication for system properties:
  - inside one SP file
  - between SP files in CLI (in builder N/A since each test case works with its own SP file)

cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/validator/CompileValidatorImpl.java
```diff
@@ -56,6 +57,18 @@ private boolean navigationListContainsKey(List<Map<String, String>> stepNavigati
         return errors;
     }
 
+    @Override
+    public void validateNoDuplicateExecutablesBasedOnFQN(
+            Executable newExecutable,
+            SlangSource newExecutableSource,
```
gituser173: newExecutableSource is not used, maybe it can be removed

cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/SlangCompilerImpl.java
```diff
@@ -54,20 +56,31 @@
...
+                Executable preCompiledCurrentSource = preCompile(currentSource);
+
+                compileValidator.validateNoDuplicateExecutablesBasedOnFQN(preCompiledCurrentSource, currentSource, allExecutables);
+
+                allExecutables.add(preCompiledCurrentSource);
             }
         }
 
-        return scoreCompiler.compile(executable, pathExecutables);
+        Set<Executable> dependencies = new HashSet<>(allExecutables);
```
gituser173: just a suggestion, maybe if you renamed allExecutables to executables than you would not need to create a new Set from which you remove the primary executable and remove it directly from the executables Set. WDYT?

cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/validator/PreCompileValidatorImpl.java
```diff
@@ -203,6 +206,38 @@ public void validateResultsSection(
         }
     }
 
+    @Override
+    public void validateNoDuplicateInputs(List<Input> inputs, Input element) {
```
gituser173: these four methods look similar, maybe they can be replaced by a generic method with a List<InOutParam> and an InOutParam as parameters

cloudslang-entities/src/main/java/io/cloudslang/lang/entities/utils/SetUtils.java
```diff
@@ -0,0 +1,64 @@
...
+import io.cloudslang.lang.entities.bindings.InOutParam;
+import java.util.Collection;
+import java.util.HashSet;
+import java.util.Set;
+
+/**
+ * @author Bonczidai Levente
+ * @since 7/6/2016
+ */
+public final class SetUtils {
```
gituser173: maybe add some unit tests for this class

gituser173:
had a few comments, otherwise it looks good :+1:

gituser173:
had a few comments, otherwise it looks good :+1:

cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/validator/CompileValidatorImpl.java
Bonczidai: that is correct

cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/SlangCompilerImpl.java
Bonczidai: alright

cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/validator/PreCompileValidatorImpl.java
Bonczidai: compiler complains because of generic collections (raw  collections), I can't call that method directly with List<SpecificType> + there are minor diffs like the error message - I think it's ok for now

cloudslang-entities/src/main/java/io/cloudslang/lang/entities/utils/SetUtils.java
Bonczidai: :+1:

cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/validator/PreCompileValidatorImpl.java
gituser173: List/<? extends InOutParam/> and InOutParam as parameters for the generic method and you could generate the method based on the second parameter of the generic method

PR #773, opened 2016-07-11
added support for sensitive data transmition between components which use different encryption key

genadi-hp:
added support for sensitive data transmition between components which use different encryption key

Sign-off: @genadi-hp, Genadi Rabinovich, genadi@hpe.com


cloudslang-runtime/src/main/java/io/cloudslang/lang/runtime/env/RunEnvironment.java
```diff
@@ -121,4 +121,103 @@ public void resetStacks() {
...
+    }
+
+    private boolean containsSensitiveReturnValues() {
+        return (returnValues != null) && (returnValues.getOutputs() != null) &&
+                containsSensitiveData(returnValues.getOutputs().values());
+    }
+
+    private boolean containsSensitiveSystemProperties() {
+        return (systemProperties != null) &&
+                containsSensitiveData(Collections2.transform(systemProperties, new Function<SystemProperty, Value>() {
```
Bonczidai: little concern: maybe we shouldn't make this class dependent of `Jython` because of a `map` functionality

PR #787, opened 2016-07-19
Input not required bug fix

gituser173:


cloudslang-runtime/src/main/java/io/cloudslang/lang/runtime/bindings/InputsBinding.java
```diff
@@ -98,14 +98,24 @@ private Value resolveValue(Input input, Map<String, ? extends Value> context, Ma
...
                 value = rawValue;
             }
         }
 
         return value;
     }
 
+    private boolean containsEmptyStringOrNull(Value value) {
+        return value != null &&
+                ((value.get() != null && value.get().equals("")) ||
```
Bonczidai: isn't this the same as `value.get() == null || value.get().equals("")` ?

Bonczidai:
:+1:

PR #802, opened 2016-07-27
Add support for creating a Value object from a pre-encrypted simple string

hprotem:


cloudslang-cli/src/test/java/io/cloudslang/lang/cli/SlangBannerTest.java
```diff
@@ -27,7 +27,14 @@ public void before() throws Exception {
 
     @Test
     public void testGetBanner() throws Exception {
-        Assert.assertEquals(BANNER, slangBanner.getBanner());
+
```
Bonczidai: note: this test is also supposed to be fixed in #795 (adding @gituser173 )

cloudslang-entities/src/main/java/io/cloudslang/lang/entities/SystemProperty.java
```diff
@@ -31,7 +31,7 @@
     private final String fullyQualifiedName;
     private final Value value;
 
-    public SystemProperty(String namespace, String key, String value, boolean sensitive) {
+    public SystemProperty(String namespace, String key, Value value) {
```
Bonczidai: this would mean we can pass any Values, however or now SPs are only strings

cloudslang-entities/src/main/java/io/cloudslang/lang/entities/bindings/values/SensitiveStringValue.java
```diff
@@ -0,0 +1,32 @@
+package io.cloudslang.lang.entities.bindings.values;
```
Bonczidai: missing licence header

PR #812, opened 2016-08-09
Support input types

### LINKED ISSUE #673: Support input types in CS
**[APPROACH I - STRINGS]** go with string interface for now, support types later

### LINKED ISSUE #797: Loop doesn't work with In range taken from system property
Steps to reproduce:
0. Create a .prop.sl file with the following property: 
properties:
  - lista: [1,2,3]
1. Run in CLI 9.64 a loop assigning the In range from the system property:
for: attempt in get_sp('tutorials_10.properties.lista')

Expected result:
The loop correctly evaluates the range and iterates through its elements

Actual result:
 Command failed java.lang.RuntimeException: Slang Error: Error running: create_em
ail_address: Error evaluating for loop expression in step 'create_email_address'
    Error is: Error in running script expression: 'get_sp('tutorials_10.properti
es.lista')',
    Exception is: Error in running script expression: 'get_sp('tutorials_10.prop
erties.lista')',
    Exception is: name 'get_sp' is not defined. Make sure to use correct syntax
for the function: get_sp('fully.qualified.name', optional_default_value).. Make
sure to use correct syntax for the function: get_sp('fully.qualified.name', opti
onal_default_value).

gituser173:
Fixes  #673, Fixes  #797


cloudslang-runtime/src/main/java/io/cloudslang/lang/runtime/env/Context.java
```diff
@@ -52,8 +52,8 @@ public void putLanguageVariable(String name, Value value) {
         langVariables.put(name, value);
     }
 
-    public Map<String, Value> getImmutableViewOfVariables() {
-        return Collections.unmodifiableMap(variables);
+    public Map<String, Value> getVariables() {
```
Bonczidai: Y

cloudslang-runtime/src/main/java/io/cloudslang/lang/runtime/env/RunEnvironment.java
```diff
@@ -169,7 +169,7 @@ private boolean containsSensitiveContexts() {
         Context context;
         while (!hasSensitive && (context = contextStack.popContext()) != null) {
             hasSensitive = containsSensitiveData(context.getImmutableViewOfLanguageVariables().values());
-            hasSensitive = hasSensitive || containsSensitiveData(context.getImmutableViewOfVariables().values());
+            hasSensitive = hasSensitive || containsSensitiveData(context.getVariables().values());
```
Bonczidai: Y

cloudslang-runtime/src/main/java/io/cloudslang/lang/runtime/env/RunEnvironment.java
```diff
@@ -210,7 +210,7 @@ public Value apply(SystemProperty systemProperty) {
         Context context;
         while ((context = contextStack.popContext()) != null) {
             valuesToCheck.addAll(context.getImmutableViewOfLanguageVariables().values());
-            valuesToCheck.addAll(context.getImmutableViewOfVariables().values());
+            valuesToCheck.addAll(context.getVariables().values());
```
Bonczidai: Y

cloudslang-runtime/src/main/java/io/cloudslang/lang/runtime/bindings/scripts/ScriptEvaluator.java
```diff
@@ -70,14 +66,21 @@ public Value evalExpr(String expr, Map<String, Value> context, Set<SystemPropert
...
+            return ValueFactory.create(validateEvalResult(result.getEvalResult()), getSensitive(result.getResultContext(), systemPropertiesDefined));
         } catch (Exception exception) {
             throw new RuntimeException(
                     "Error in running script expression: '"
                             + expr + "',\n\tException is: " + handleExceptionSpecialCases(exception.getMessage()), exception);
         }
     }
 
+    private Serializable validateEvalResult(Serializable evalResult) {
+        if (evalResult != null && !(evalResult instanceof String)) {
```
Bonczidai: can we reduce code duplication?

cloudslang-runtime/src/main/java/io/cloudslang/lang/runtime/bindings/ResultsBinding.java
```diff
@@ -102,12 +103,13 @@ public String resolveResult(Map<String, Value> inputs,
                 try {
                     Value expressionResult = scriptEvaluator.evalExpr(expression, scriptContext, systemProperties, result.getFunctionDependencies());
                     Boolean evaluatedResult;
-                    if (expressionResult.get() instanceof Integer) {
-                        evaluatedResult = (Integer) expressionResult.get() != 0;
+                    String expressionResultValue = (String) expressionResult.get();
+                    if (NumberUtils.isNumber(expressionResultValue)) {
```
Bonczidai: I think in results case, the script eval should return a boolean as before (no change), typing is only for variables

Bonczidai: so you need to move a bit the validation in ScriptEvaluator

Bonczidai:
@gituser173 it looks good :+1: 

- I don't think we should remove `getImmutableView..`, please investigate a bit the situation, how come it worked before?
- you should also add `private Set<String> systemPropertyDependencies;` to the loop statement and add corresponding handling, check the useage of it in `InOutParam` class; basically `CompilationArtifact`'s SP dependencies skips SP from loop statements (please add test for this -- check the usage of `io.cloudslang.lang.entities.CompilationArtifact#getSystemProperties`)
- be sure to have negative tests, for input types other than string

Bonczidai:
much better :+1: 

- did you handle the second comment? 

```
you should also add private Set<String> systemPropertyDependencies; to the loop statement and add corresponding handling, check the usage of it in InOutParam class; basically CompilationArtifact's SP dependencies skips SP from loop statements (please add test for this -- check the usage of io.cloudslang.lang.entities.CompilationArtifact#getSystemProperties)
```

you can also heck out `io.cloudslang.lang.compiler.modeller.DependenciesHelper#getSystemPropertiesForFlow` to see how we calculate it for flow, which is also your case

- please check out why the build is failing

Bonczidai:
is the SP dependency stuff working? did you add the test? in the compiler ~ `io.cloudslang.lang.systemtests.flows.FunctionDependenciesTest#testSystemPropertyDependencies`

Bonczidai:
looks good :+1:

Bonczidai:
you can merge it once button appears

PR #813, opened 2016-08-09
Step type fixes

### LINKED ISSUE #811: Request: make LanguageEventData.getStepType more specific
We have a defect in OO whereby in the table view of the run drilldown the icons denoting step types are not shown. The reason is that LanguageEventData.getStepType is not specific enough. For all of the following types: [operation, sub-flow, decision] it returns LanguageEventData.StepType#EXECUTABLE.

Can we replace StepType#EXECUTABLE by 3 other types: StepType#OPERATION, StepType#DECISION, StepType#FLOW? Or is there any other way to detect the exact step type?

Bonczidai:
fixes #811 


Bonczidai:
note:

migrated all the events related stuff to LangaugeEventData, now an event can be of type:
```
        FLOW("flow"),
        OPERATION("operation"),
        DECISION("decision"),
        STEP("step"),
        ACTION("action"),
        NAVIGATION("navigation");
```

p.s. do not confuse this with `io.cloudslang.lang.entities.ExecutableType` which is used throughout the langauge (there is a method in LanguageEventData that does the conversion)

PR #814, opened 2016-08-10
Result changes, validations

### LINKED ISSUE #800: Operation\Decision with no results should have only Success by default
Operation that has no results gets Default Success and Failure
But Failure can never occur if Success is added with no expression (it is always true)
This means the Operation has really only Success.

The client (for example in OOD) that uses this operation, sees that it has Failure and trusts that some kind of failure was handled when it is actuality did not. It is more likely to assume that this operation did not get default results, but was handled with both success expression and failure expression.

Since the Failure is unreachable, there should be only default Success.

### LINKED ISSUE #806: Validate in CS  behavior of results of operation and desicion
do we validate we always have default \ "true" result?

Bonczidai:
fixes #806 
fixes #800 

### DIFF
--- a/cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/modeller/ExecutableBuilder.java
+++ b/cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/modeller/ExecutableBuilder.java
@@ -8,7 +8,6 @@
  */
 package io.cloudslang.lang.compiler.modeller;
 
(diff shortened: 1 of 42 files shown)


Bonczidai:
#806: results section can contain multiple expression results and a default ending result (only one, mandatory). Default result can be: result without values (`-result`) or with a boolean true value (`-result: true`).

#800: by default flows have S+F, ops have S, decisions have none

PR #818, opened 2016-08-17
validate navigation section against dependency results

### LINKED ISSUE #808: Validate step declares exact dependency results in navigation section
missing navigation in step means that it has default behavior.
in case the original step results does not contain failure and success (only the two of them) we want the compilation to fail.

Bonczidai:
Fixes #808 

### DIFF
--- a/cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/validator/CompileValidatorImpl.java
+++ b/cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/validator/CompileValidatorImpl.java
@@ -149,7 +149,7 @@ public class CompileValidatorImpl extends AbstractValidator implements CompileVa
     }
 
     private void validateResultNamesAndNavigationSection(Flow flow, Step step, String refId, Executable reference, List<RuntimeException> errors) {
-        List<String> stepNavigationKeys = mapMapToKey(step.getNavigationStrings());
+        List<String> stepNavigationKeys = getMapKeyList(step.getNavigationStrings());
         List<String> refResults = getReferenceResultNames(step, reference.getResults());
 
         List<String> stepNavigationKeysWithoutMatchingResult = ListUtils.subtract(stepNavigationKeys, refResults);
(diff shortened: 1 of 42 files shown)


cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/validator/CompileValidatorImpl.java
```diff
@@ -138,28 +133,73 @@ private boolean isForLoop(Step step, List<String> breakValuesList) {
...
+            ));
+        }
+    }
+
+    private String getErrorMessagePrefix(Flow flow, Step step) {
+        return "Cannot compile flow '" + flow.getName() +
+                "' since for step '" + step.getName() + "'";
+    }
+
+    private List<String> mapMapToKey(List<Map<String, String>> collection) {
```
gituser173: maybe rename to "getKeyList" or something better

cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/validator/CompileValidatorImpl.java
```diff
@@ -138,28 +133,73 @@ private boolean isForLoop(Step step, List<String> breakValuesList) {
...
+        List<String> result = new ArrayList<>();
+        for (Map<String, String> element : collection) {
+            result.add(element.keySet().iterator().next());
+        }
+        return result;
+    }
+
+    private List<String> mapResultsToNames(List<Result> results) {
```
gituser173: "getResultsNamesList"? WDYT?

gituser173:
Looks good :+1:

cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/validator/CompileValidatorImpl.java
Bonczidai: alright

cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/validator/CompileValidatorImpl.java
Bonczidai: that's also perfectly valid - but this method is actually a clear lambda expression (in python it would be ~ `map(lambda x : x['name'], results))`, so I would like to stay with this naming (it will be easier to spot when we move to Java 8)

Bonczidai:
@gituser173 thanks for the review, merging

PR #819, opened 2016-08-17
validate result name against on_failure

### LINKED ISSUE #796: Step/Result should not accept name on failure
Step/result cannot be called on-failure

Step: “Below 'on_failure' property there should be a list of steps and not a map” – this error message should be improved
Result: the check does not work

Expected:
No step or response should be called 'on_failure' or any other case combination e.g.: 'ON_Failure', 'ON_FAILURE' etc 
Relevant message should be displayed at compilation

Bonczidai:
fixes #796 

### DIFF
--- a/cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/validator/PreCompileValidatorImpl.java
+++ b/cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/validator/PreCompileValidatorImpl.java
@@ -283,6 +283,17 @@ public class PreCompileValidatorImpl extends AbstractValidator implements PreCom
         }
     }
 
+    @Override
+    public void validateResultName(String resultName) {
+        validateKeywords(resultName);
+    }
+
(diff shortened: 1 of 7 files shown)


cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/validator/PreCompileValidatorImpl.java
```diff
@@ -283,6 +283,17 @@ public void validateDefaultResult(List<Result> results, String artifactName, Lis
         }
     }
 
+    @Override
+    public void validateResultName(String resultName) {
+        validateKeywords(resultName);
+    }
+
+    private void validateKeywords(String resultName) {
+        if (SlangTextualKeys.ON_FAILURE_KEY.equalsIgnoreCase(resultName)) {
```
gituser173: don't we need to validate against more keywords?

gituser173:
Looks good :+1:

PR #823, opened 2016-08-24
Characters convention

### LINKED ISSUE #817: Characters convention
Characters convention in names of input/output & flow/operation/decision+instantiating steps

- **Which are the permitted / illegal characters in input/output names?** 

We should have a documented convention on this ( as for [namespaces (project & folder)](http://cloudslang-docs.readthedocs.io/en/v0.9.60/cloudslang_best_practices.html) and [system properties](https://github.com/CloudSlang/cloud-slang/pull/815) - \w-. )

A discussion was started in https://github.com/CloudSlang/docs/issues/212, but "In general, CloudSlang variable names must conform to both Python’s naming constraints as well as Java’s naming constraints." is not explicit for users/applications writing CS code.
     Does this mean only letters, digits and _ , starting with a letter ?
Also do we have a list of (java/python/yaml) keywords which are forbidden to use?

- The same problem should be clarified for the **names of results & flows,operation,decisions+the steps instantiating them**

Bonczidai:
fixes #817


Bonczidai:
@SamMarkowitz we need to document these validations

cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/SlangCompilerImpl.java
```diff
@@ -102,10 +102,14 @@ public ExecutableModellingResult preCompileSource(SlangSource source) {
...
         //first thing we parse the yaml file into java maps
-        ParsedSlang parsedSlang = yamlParser.parse(source);
+        ParseModellingResult parseModellingResult = yamlParser.parseAndValidate(source);
 
         // Then we transform the parsed Slang source to a Slang model
-        return slangModeller.createModel(parsedSlang);
+        ExecutableModellingResult executableModellingResult =
+                slangModeller.createModel(parseModellingResult.getParsedSlang());
+
+        executableModellingResult.getErrors().addAll(parseModellingResult.getErrors());
```
gituser173: maybe you could move the above line in the slangModeller.createModel method, WDYT?

cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/validator/ExecutableValidatorImpl.java
```diff
@@ -0,0 +1,149 @@
...
+        ParsedSlang.Type executableType = parsedSlang.getType();
+        switch (executableType) {
+            case SYSTEM_PROPERTY_FILE:
+                systemPropertyValidator.validateNamespace(namespace);
+                break;
+            case FLOW:
+            case OPERATION:
+            case DECISION:
+                // namespace cannot be empty
+                if (org.apache.commons.lang3.StringUtils.isEmpty(namespace)) {
```
gituser173: why are you using two different versions of StringUtils?

cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/modeller/ExecutableBuilder.java
```diff
@@ -437,9 +438,20 @@ private WorkflowModellingResult compileWorkFlow(List<Map<String, Map<String, Obj
         return new WorkflowModellingResult(new Workflow(steps), errors);
     }
 
+    private String getStepName(Map<String, Map<String, Object>> stepRawData, List<RuntimeException> errors) {
+        String stepName = null;
+        try {
+            stepName = stepRawData.keySet().iterator().next();
+            executableValidator.validateStepName(stepName);
```
gituser173: maybe we should have a convention for validators that all of them throw exceptions or all of them have an errors list as a parameter which is returned by the validators, WDYT?

cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/modeller/ExecutableBuilder.java
```diff
@@ -478,11 +490,15 @@ private StepModellingResult compileStep(
...
             errors.add(new RuntimeException("Step: \'" + stepName + "\' has no reference information"));
         } else {
-            String refString = doRawData.keySet().iterator().next();
-            refId = resolveReferenceID(refString, imports, namespace);
+            try {
+                String refString = doRawData.keySet().iterator().next();
+                refId = resolveReferenceID(refString, imports, namespace);
+            } catch (RuntimeException rex) {
+                errors.add(rex);
+            }
```
gituser173: extractMethod maybe?

cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/modeller/ExecutableBuilder.java
```diff
@@ -478,11 +490,15 @@ private StepModellingResult compileStep(
         if (MapUtils.isEmpty(doRawData)) {
             errors.add(new RuntimeException("Step: \'" + stepName + "\' has no reference information"));
         } else {
-            String refString = doRawData.keySet().iterator().next();
-            refId = resolveReferenceID(refString, imports, namespace);
+            try {
+                String refString = doRawData.keySet().iterator().next();
+                refId = resolveReferenceID(refString, imports, namespace);
+            } catch (RuntimeException rex) {
+                errors.add(rex);
```
gituser173: if all the validators would return a list of exceptions we would not need to do this, but we need to decide on a convention

cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/modeller/transformers/AbstractForTransformer.java
```diff
@@ -105,7 +84,33 @@
...
+                    dependencyAccumulator.getSystemPropertyDependencies());
+        } else {
+            loopStatement = new ListForLoopStatement(varName, collectionExpression,
+                    dependencyAccumulator.getFunctionDependencies(),
+                    dependencyAccumulator.getSystemPropertyDependencies());
+        }
+        return loopStatement;
+    }
+
+    private void validateChars(String name) {
```
gituser173: do we need this small method? seems reduntant

cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/modeller/transformers/BreakTransformer.java
```diff
@@ -29,7 +33,12 @@
         if (rawData == null) {
             transformedData.add(ScoreLangConstants.FAILURE_RESULT);
         } else {
-            transformedData = rawData;
+            try {
+                executableValidator.validateBreakKeys(rawData);
```
gituser173: same comment about these validations methods

cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/validator/PreCompileValidatorImpl.java
```diff
@@ -308,9 +300,17 @@ private String getMessagePart(Class aClass) {
...
 
-    private String getExecutableName(Map<String, Object> executableRawData) {
+    private String getExecutableName(Map<String, Object> executableRawData, List<RuntimeException> errors) {
         String execName = (String) executableRawData.get(SlangTextualKeys.EXECUTABLE_NAME_KEY);
-        return execName == null ? "" : execName;
+        if (StringUtils.isBlank(execName)) {
+            errors.add(new RuntimeException("Executable has no name"));
+        }
+        try {
+            executableValidator.validateExecutableName(execName);
```
gituser173: you could remove the dependency on ExecutableValidator and use the validateSimpleNameRules from AbstractValidator

gituser173:
:+1: nice work, had a few comments

cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/SlangCompilerImpl.java
Bonczidai: :+1:

cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/validator/ExecutableValidatorImpl.java
Bonczidai: good catch

cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/modeller/ExecutableBuilder.java
lucian-cm: getStepName has side effects. When calling the method two times with the same parameters makes the list populated twice with error message, a get method should be idempotent. Th way to go here is to have a metod validateStepName that validates a stepName and a separate getStepName method.

cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/SlangCompilerImpl.java
```diff
@@ -102,10 +102,10 @@ public ExecutableModellingResult preCompileSource(SlangSource source) {
         Validate.notNull(source, "You must supply a source to compile");
 
         //first thing we parse the yaml file into java maps
-        ParsedSlang parsedSlang = yamlParser.parse(source);
```
lucian-cm: it is a little not intuitive to have a method called `parseAndValidate`, a method should do only one thing in general. Can it be that parse and validate can be separated without introducing performance problems? If that is not possible maybe a better name is `validate` as it is already called on a yamlParser

cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/SlangCompilerImpl.java
```diff
@@ -127,7 +127,7 @@ public ExecutableModellingResult preCompileSource(SlangSource source) {
     }
 
     private ParsedSlang parseSystemPropertiesFile(SlangSource source) {
-        ParsedSlang parsedSlang = yamlParser.parse(source);
+        ParsedSlang parsedSlang = yamlParser.parseAndValidateAndThrowFirstError(source);
```
lucian-cm: seems that different it would be more generic to be able to wire in the interface method a strategy to validate, for example proceed_on_failure (this strategy collects all validation errors and message), fail_on_first (thorws exception on first invalid thing)

cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/modeller/ExecutableBuilder.java
```diff
@@ -50,6 +41,15 @@
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import org.apache.commons.collections4.CollectionUtils;
+import org.apache.commons.collections4.ListUtils;
+import org.apache.commons.collections4.MapUtils;
+import org.apache.commons.collections4.iterators.PeekingIterator;
+import org.apache.commons.lang.StringUtils;
```
lucian-cm: lang3

cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/validator/matcher/PatternMatcher.java
```diff
@@ -0,0 +1,27 @@
...
+ * @since 8/30/2016
+ */
+public abstract class PatternMatcher {
+    private final Pattern pattern;
+
+    public PatternMatcher(Pattern pattern) {
+        this.pattern = pattern;
+    }
+
+    public boolean matchesEndToEnd(String input) {
```
lucian-cm: `matches `is a good name, following the Java convention, I should say that I don't necessarily am a fan of wrapping `Pattern`s. The typical use of a Pattern is to compile the regex once and then reuse it. Maybe check if we do need this dynamic `Pattern`s and if not replace with static `Pattern`s.

cloudslang-compiler/src/test/java/io/cloudslang/lang/compiler/CharValidationsPreCompileErrorsTest.java
```diff
@@ -0,0 +1,142 @@
...
+
+import static org.junit.Assert.assertTrue;
+
+/**
+ * @author Bonczidai Levente
+ * @since 8/26/2016
+ */
+@RunWith(SpringJUnit4ClassRunner.class)
+@ContextConfiguration(classes = SlangCompilerSpringConfig.class)
+public class CharValidationsPreCompileErrorsTest {
```
lucian-cm: this test is great, but mock testing is a bit faster

lucian-cm:
@Bonczidai great work, had a few comments :+1:

cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/modeller/ExecutableBuilder.java
Bonczidai: :+1: @lucian-cm 

@gituser173 agree, I'll open an issue on that

cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/modeller/ExecutableBuilder.java
Bonczidai: :+1: https://github.com/CloudSlang/cloud-slang/issues/849

cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/modeller/ExecutableBuilder.java
Bonczidai: let's wait for validator convention: https://github.com/CloudSlang/cloud-slang/issues/849

cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/modeller/transformers/AbstractForTransformer.java
Bonczidai: :+1:

cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/modeller/transformers/BreakTransformer.java
Bonczidai: https://github.com/CloudSlang/cloud-slang/issues/849

cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/validator/PreCompileValidatorImpl.java
Bonczidai: I think it's better this way, since ExecutableValidator has the specific validate methods, whereas AbstractValidator has general logic, in the future we may want to aplpy other validation rules to exec name than `simpleNameRules`

cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/SlangCompilerImpl.java
Bonczidai: :+1:

cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/modeller/ExecutableBuilder.java
Bonczidai: :+1:

cloudslang-compiler/src/test/java/io/cloudslang/lang/compiler/CharValidationsPreCompileErrorsTest.java
Bonczidai: agree, but the tests on top level package (`io.cloudslang.lang.compiler`) are meant to work with real objects - the other sub-packages are tested with mocks

Bonczidai:
thanks @gituser173 and @lucian-cm for the review

PR #850, opened 2016-09-08
Flow input which is required & private and has a value of empty strin…

### LINKED ISSUE #843: Flow input which is required & private and has a value of empty string should be invalid

gituser173:
Fixes #843


cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/modeller/transformers/AbstractInputsTransformer.java
```diff
@@ -93,7 +93,10 @@ private Input createPropInput(Map.Entry<String, Map<String, Serializable>> entry
 
         if (privateInput && !defaultSpecified) {
             throw new RuntimeException(
-                    "Input: " + inputName + " is private but no default value was specified");
+                    "Input: '" + inputName + "' is private but no default value was specified");
```
Bonczidai: I think we should merge these two if branches into one condition

the defaultSpecified should cover ~ StringUtils.isNotEmpty(proper.get(key))

currently
```
    - input1:
        default: None
        private: true
        required: true
```

case is not covered - please add a test

Bonczidai:
maybe we can merge first #851 to reuse InputUtils / ValueUtils here

Bonczidai:
👍

PR #855, opened 2016-09-12
When a step has loop and the for statement is empty, it passes CS compilation. Should fail

### LINKED ISSUE #841: When a step has loop and the for statement is empty, it passes CS compilation. Should fail

gituser173:
Fixes #841

### DIFF
--- a/cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/modeller/TransformersHandler.java
+++ b/cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/modeller/TransformersHandler.java
@@ -54,8 +54,10 @@ public class TransformersHandler {
                 if (data != null) {
                     transformedData.put(key, (Serializable) data);
                 }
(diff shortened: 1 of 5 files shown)


cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/modeller/TransformersHandler.java
```diff
@@ -54,8 +54,10 @@ public static String keyToTransform(Transformer transformer) {
                 if (data != null) {
                     transformedData.put(key, (Serializable) data);
                 }
-                for (RuntimeException rex : transformModellingResult.getErrors()) {
-                    errors.add(wrapErrorMessage(rex, errorMessagePrefix));
+                if (rawData.containsKey(key)) {
```
Bonczidai: hmz

Bonczidai:
👍 please also add a test for `for: ''` case

Bonczidai:
🆒

PR #859, opened 2016-09-14
System properties are not retrived in case duplicate ids 

### LINKED ISSUE #848: System properties are not retrived in case duplicate ids
Currently, when a SP file has any error, like duplicate identifier, the precompile does not retrieve the SPs inside.
This mean that in WS if I have a file SP1 with two identifiers SP1.a and SP1.a and I create in a new file SP2 with SP1.a identifier I will not get it as duplicate.

I think SP precompile should behave like a flow and return the content of the file.

Possible solution:

We can separate the layers like we do with precompile: preCompileSource -> model (what we could extract) + exceptions; precompile -> model or throw first exception,
We can do something similar for SPs

gituser173:
Fixes #848

### DIFF
--- a/cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/SlangCompilerImpl.java
+++ b/cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/SlangCompilerImpl.java
@@ -12,6 +12,7 @@ import io.cloudslang.lang.compiler.modeller.SlangModeller;
 import io.cloudslang.lang.compiler.modeller.model.Executable;
 import io.cloudslang.lang.compiler.modeller.result.ExecutableModellingResult;
 import io.cloudslang.lang.compiler.modeller.result.ParseModellingResult;
+import io.cloudslang.lang.compiler.modeller.result.SystemPropertyModellingResult;
 import io.cloudslang.lang.compiler.parser.YamlParser;
 import io.cloudslang.lang.compiler.parser.model.ParsedSlang;
 import io.cloudslang.lang.compiler.scorecompiler.ScoreCompiler;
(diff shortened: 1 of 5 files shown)


cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/SlangCompilerImpl.java
```diff
@@ -116,105 +117,130 @@ public ExecutableModellingResult preCompileSource(SlangSource source) {
...
+        }
+        return systemPropertyModellingResult.getSystemProperties();
+    }
+
+    @Override
+    public SystemPropertyModellingResult loadSystemPropertiesFromSource(SlangSource source) {
+        ParseModellingResult parseModellingResult = parseSystemPropertiesFile(source);
+        SystemPropertyModellingResult systemPropertyModellingResult =
+                extractProperties(parseModellingResult.getParsedSlang(), source);
+        systemPropertyModellingResult.getErrors().addAll(parseModellingResult.getErrors());
```
Bonczidai: I'd put the parse errors before the extract properties errors

cloudslang-compiler/src/test/java/io/cloudslang/lang/compiler/LoadSystemPropertiesTest.java
```diff
@@ -183,6 +210,19 @@ public void testInvalidCharsNamespace() throws Exception {
     }
 
     @Test
+    public void testInvalidCharsNamespaceFromSource() throws Exception {
```
Bonczidai: IMHO you should validate that you get all the exceptions when having multiple ones

Bonczidai:
👍 take a look on the comments

PR #864, opened 2016-09-15
Cs line errors

lucian-cm:
Making more developer friendly to use cloud slang compiler when using without Spring support


lucian-cm:
@Bonczidai please review

Bonczidai:
👍 💪

PR #880, opened 2016-09-27
CloudSlang build tool should print for each test case name - from which file is it

Bonczidai:
fixes #249 

Signed-off-by: Bonczidai Levente <levente.bonczidai@hpe.com>

PR #892, opened 2016-10-04
CloudSlang build tool should print for each test case name - from which file is it

### LINKED ISSUE #249: CloudSlang build tool should print for each test case name - from which file is it

Bonczidai:
fixes #249, closes #880 


Bonczidai:
TS:

SlangSource changes - it also stores the files path. Mandatory fields (content, name), optional fields (file path, extension)
store file path in SlangTestCase

Issue #890, opened 2016-10-03
[Build] Used but undeclared Maven dependencies do not fail the build of a module


lucian-cm:
Currently in some module there are some dependencies that are used in production code, that are transitive dependencies of other modules or other third parties and not declared directly. This is a bad practice and we should be protected from such cases at build time.

PR #900, opened 2016-10-05
Adding dependency checker for all cloudslang modules

lucian-cm:
Fix for #890
Introduce a mechanism to fail the build in case of undeclared dependencies which are used.
Also introduced a mechanism to fail the build in case of used but declared. This might produce several false positives so for now we will mostly investigate such findings adding them in the ignore list.

### DIFF
--- a/cloudslang-all/pom.xml
+++ b/cloudslang-all/pom.xml
@@ -55,5 +55,37 @@ http://www.apache.org/licenses/LICENSE-2.0
 
     </dependencies>
 
+    <build>
+        <plugins>
+
+            <plugin>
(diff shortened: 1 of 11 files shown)


Bonczidai:
LGTM - right now the plugin is only triggered for `cloudslang-content-verifier` and `cloudslang-cli` modules, I think we should run it for all of them

lucian-cm:
@Bonczidai Fixing the rest of modules.
Right now we have this disabled for cloudslang-tests until we fix the problems in that module, since in there we have to use 2.8 version that does no support such customization options like 2.8
All issue reported in the ignore list need to be fixed and I opened an issue for that #902

PR #901, opened 2016-10-05
Adding dependency checker for all cloudslang modules

lucian-cm:
Fix for #890 
Introduce a mechanism to fail the build in case of undeclared dependencies which are used.
Also introduced a mechanism to fail the build in case of used but declared. This might produce several false positives so for now we will mostly investigate such findings adding them in the ignore list.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -489,6 +489,62 @@
                         </execution>
                     </executions>
                 </plugin>
+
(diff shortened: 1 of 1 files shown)


Bonczidai:
we have code freeze on `1.0`, I think it' enough to handle this on master branch

👌

PR #946, opened 2016-11-04
Spring cleansing from 3 modules

### LINKED ISSUE #894: [IntelliJ Plugin] Remove Spring compile time traces in modules cloudslang-compiler, cloudslang-entities, cloudslang-spi
The motivation for this is to maintain a single code base for IntelliJ Plugin and the code in the official cloud-slang repo.
The ultimate goal is to remove Spring dependencies (spring-core, spring-context, spring-beans, spring-aop etc) from the IntelliJ Plugin. Other refactoring solutions of the module should be considered if possible.

lucian-cm:
Fixes #894
'Removing' Spring to have 1 codebase between plugin and cloud slang modules


cloudslang-entities/src/main/java/configuration/SlangEntitiesSpringConfig.java
```diff
@@ -0,0 +1,48 @@
...
+public class SlangEntitiesSpringConfig implements ApplicationContextAware {
+
+    private static ApplicationContext applicationContext;
+
+    public static Encryption[] getEncryptors() {
+        if (applicationContext != null) {
+            Map<String, Encryption> encryptorMap = applicationContext.getBeansOfType(Encryption.class);
+            return encryptorMap.values().toArray(new Encryption[encryptorMap.size()]);
+        } else {
+            throw new RuntimeException("ApplicationContextProvider bean missing");
```
Bonczidai: IMHO in this case we should use `ApplicationContext bean missing` message - and extract it as const (it also used in `EncryptionProvider`)

cloudslang-entities/src/main/java/io/cloudslang/lang/entities/encryption/EncryptionProvider.java
```diff
@@ -34,21 +34,25 @@ public static Encryption get() {
...
-                return encryptors[0];
-            } else if (encryptors.length == 2) {
-                return encryptors[0] instanceof DummyEncryptor ? encryptors[1] : encryptors[0];
-            } else {
-                throw new RuntimeException("Too many (" + encryptors.length + ") encryptors found");
+        Encryption[] encryptors;
+        try {
+            encryptors = SlangEntitiesSpringConfig.getEncryptors();
+        } catch (Exception theGivenEx) {
+            if (!StringUtils.containsIgnoreCase(theGivenEx.getMessage(), "ApplicationContextProvider bean missing")) {
```
Bonczidai: hmm - do we need the negation here?

cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/modeller/ExecutableBuilder.java
```diff
@@ -106,7 +95,6 @@
             asList(ScoreLangConstants.LOOP_KEY, SlangTextualKeys.DO_KEY, SlangTextualKeys.NAVIGATION_KEY);
     private List<String> parallelLoopValidKeywords = asList(SlangTextualKeys.DO_KEY, SlangTextualKeys.FOR_KEY);
 
-    @PostConstruct
```
Bonczidai: maybe we should leave this as a comment to mark the logic or switch to a builder / constructor approach to call this method in the constructor after all the setters - WDYT? Is it good enough to have the init logic only in spring config class?

cloudslang-entities/src/main/java/configuration/SlangEntitiesSpringConfig.java
lucian-cm: Nice catch 👍  Please see last commit.

cloudslang-entities/src/main/java/io/cloudslang/lang/entities/encryption/EncryptionProvider.java
lucian-cm: I don't like negations as well, but in this case it is expressing the intention of the code 👍

cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/modeller/ExecutableBuilder.java
lucian-cm: Added comment to mark it as post construction method

PR #949, opened 2016-11-06
Builder log4j config

### LINKED ISSUE #948: Log4j not working properly for builder
should configure it from `configuration` folder externally

Bonczidai:
fixes #948

### DIFF
--- a/cloudslang-content-verifier/src/main/java/io/cloudslang/lang/tools/build/SlangBuildMain.java
+++ b/cloudslang-content-verifier/src/main/java/io/cloudslang/lang/tools/build/SlangBuildMain.java
@@ -104,6 +104,8 @@ public class SlangBuildMain {
     private static final String MESSAGE_ERROR_LOADING_SMART_MODE_CONFIG_FILE = "Error loading smart " +
             "mode configuration file:";
     private static final String LOG4J_CONFIGURATION_KEY = "log4j.configuration";
+    private static final String LOG4J_ERROR_PREFIX = "log4j: error loading log4j properties file.";
+    private static final String LOG4J_ERROR_SUFFIX = "Using default configuration.";
 
 
     // This class is a used in the interaction with the run configuration property file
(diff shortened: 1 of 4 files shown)


cloudslang-content-verifier/src/main/java/io/cloudslang/lang/tools/build/SlangBuildMain.java
```diff
@@ -282,6 +288,33 @@ public static void main(String[] args) {
...
 
+    private static void configureLog4j() {
+        String configFilename = System.getProperty(LOG4J_CONFIGURATION_KEY);
+        String errorMessage = null;
+
+        if (StringUtils.isEmpty(configFilename)) {
+            errorMessage = "Config file name is empty.";
+        } else {
+            String normalizedPath = FilenameUtils.normalize(configFilename);
+            if (!get(normalizedPath).isAbsolute()) {
```
Bonczidai: @lucian-cm do you think from security perspective we should only accept absolute paths?

lucian-cm: The `log4j.properties` path must be constrained to be inside the `verifier `root folder. It is not that important if it is a relative path or absolute path as they are both subject to directory traversal attack, but it is a tad easier to refer to something inside the verifier root folder using relative paths.
We must make sure that we disallow special sequences like ".." and "." inside path and this verification should be done without file access api like `Paths `or `File`.
https://cwe.mitre.org/data/definitions/22.html

Given all these, http://grepcode.com/file/repo1.maven.org/maven2/commons-io/commons-io/1.4/org/apache/commons/io/FilenameUtils.java#FilenameUtils.normalize%28java.lang.String%29 `FileNameUtils.normalize` handles `./` and `../`
but as it shows in the documentation we should check for null afterwards, as multiple parent calls can yield that. So in conclusion we should just add the null check after normalization and the validation that the path starts with the verifier folder using startsWith on normalized strings. Then it is safe to access the file using File and Paths API. 

I would use as a preference the `PropertyConfigurer.configure(Properties properties)` because it gives far less headaches and we can give an extra error related to the file failing verification of being a property file.

Bonczidai:
@lucian-cm please check

Issue #947, opened 2016-11-05
Create Integration tests for CloudSlang CLI and builder


lucian-cm:
There is a need to have integration testing for the CLI and builder to have a higher degree of confidence in big refactorings

PR #953, opened 2016-11-10
Fix score compile with error case

### LINKED ISSUE #952: Builder does not fail
at `validate against dependencies phase`

Bonczidai:
fixes #952 
fixes CloudSlang/cloud-slang-content#846


Bonczidai:
tests need to be added in #947 - test that builder, CLI and the the CS compilation API fails when step is not correct related to its dependency

PR #988, opened 2017-05-05
execution plan bug fix

gituser173:

### DIFF
--- a/cloudslang-entities/src/main/java/io/cloudslang/lang/entities/bindings/values/SensitiveValue.java
+++ b/cloudslang-entities/src/main/java/io/cloudslang/lang/entities/bindings/values/SensitiveValue.java
@@ -11,15 +11,17 @@ package io.cloudslang.lang.entities.bindings.values;
 
 import com.fasterxml.jackson.annotation.JsonIgnore;
 import io.cloudslang.lang.entities.encryption.EncryptionProvider;
+import io.cloudslang.lang.spi.encryption.Encryption;
+import javassist.util.proxy.ProxyObjectInputStream;
+import javassist.util.proxy.ProxyObjectOutputStream;
+import org.apache.commons.codec.binary.Base64;
+import org.apache.commons.io.IOUtils;
+
 import java.io.ByteArrayInputStream;
 import java.io.ByteArrayOutputStream;
 import java.io.ObjectInputStream;
 import java.io.ObjectOutputStream;
 import java.io.Serializable;
-import javassist.util.proxy.ProxyObjectInputStream;
-import javassist.util.proxy.ProxyObjectOutputStream;
-import org.apache.commons.codec.binary.Base64;
-import org.apache.commons.io.IOUtils;
 
 /**
  * Sensitive InOutParam value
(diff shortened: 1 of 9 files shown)


cloudslang-entities/src/main/java/io/cloudslang/lang/entities/bindings/values/SensitiveValue.java
```diff
@@ -29,6 +31,7 @@
 public class SensitiveValue implements Value {
 
     public static final String SENSITIVE_VALUE_MASK = "********";
+    public static final String ENCRYPTOR_PACKAGE = "com.hp.onyx.commons.encryption";
```
Bonczidai: I still don't think we should put company related info in open source and make CS dependent on OO codebase - I would rather just pass a single constructor param to set the mode -- encryption / obfuscation or maybe a System Property -- that way we could easily add unit tests at CS level -- if somebody wants to modify the obfuscation / enc behavior he can do it anyway

@lucian-cm WDYT?

cloudslang-entities/src/main/java/io/cloudslang/lang/entities/encryption/DummyEncryptor.java
```diff
@@ -19,6 +19,11 @@ public String encrypt(char[] clearText) {
     }
 
     @Override
+    public String obfuscate(String cypherText) {
```
gituser173: no, the decrypt method does deObfuscation and decrypt

cloudslang-entities/src/main/java/io/cloudslang/lang/entities/bindings/values/SensitiveValue.java
```diff
@@ -40,6 +43,8 @@
     @JsonIgnore
     private Serializable originalContent = null;
 
+    private boolean isExecutionPlan;
```
gituser173: this value is false by default which is the case for all other constructors
@lucian-cm WDYT?

lucian-cm:
@gituser173 , as a general comment, I fear that the method modified io.cloudslang.lang.compiler.modeller.transformers.AbstractInputsTransformer#createInput(java.lang.String, java.io.Serializable, boolean, boolean, boolean) is actually used from several points in the CloudSlang codebase.

This makes this pull request a bit unsafe and probably not equivalent.
I agree with @Bonczidai comment about the package.

What I think the best solution is for this case is to make an entry point method for the specific compile service from where you pass as an enum the type of data operation to be done (for encryption or obfuscation) and this ensures that CloudSlang code remains the same. And these removes the need to put that specific package reference.

cloudslang-entities/src/main/java/io/cloudslang/lang/entities/bindings/values/SensitiveValue.java
lucian-cm: @Bonczidai totally agree

cloudslang-entities/src/main/java/io/cloudslang/lang/entities/bindings/values/SensitiveValue.java
lucian-cm: I prefer constructors as well, but surely an = initializer will work as well.

PR #989, opened 2017-05-08
Added obfuscation and deobfuscation to inputs

### LINKED ISSUE #990: Add obfuscation and deobfuscation to inputs

gituser173:
Fixes #990 

### DIFF
--- a/cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/SlangCompiler.java
+++ b/cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/SlangCompiler.java
@@ -41,6 +41,12 @@ public interface SlangCompiler {
      */
     CompilationArtifact compile(SlangSource source, Set<SlangSource> path, PrecompileStrategy precompileStrategy);
 
+    CompilationArtifact compile(
+            SlangSource source,
+            Set<SlangSource> path,
+            PrecompileStrategy precompileStrategy,
+            SensitivityLevel sensitivityLevel);
+
     /**
      * Compile a CloudSlang source and its dependencies to a
      *     {@link io.cloudslang.lang.entities.CompilationArtifact} object
@@ -107,6 +113,18 @@ public interface SlangCompiler {
      *     or an operations in the file, and a list of all the errors that were found (if any).
      */
     ExecutableModellingResult preCompileSource(SlangSource source, PrecompileStrategy precompileStrategy);
+
+    /**
+     * Pre-compile a CloudSlang source into an {@link ExecutableModellingResult}.
+     * All errors that are found during pre-compilation are collected (an exception is not thrown)
+     *
+     * @param source             the {@link SlangSource}
+     * @param precompileStrategy whether to use caching in pre-compile.
+     * @param sensitivityLevel
+     * @return an {@link ExecutableModellingResult} object, containing an executable which is either a flow
+     *     or an operations in the file, and a list of all the errors that were found (if any).
+     */
+    ExecutableModellingResult preCompileSource(SlangSource source, PrecompileStrategy precompileStrategy, SensitivityLevel sensitivityLevel);
 
     /**
      * Remove all elements in pre-compile cache. No-cached calls are not affected.
--- a/cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/SlangCompiler.java
+++ b/cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/SlangCompiler.java
@@ -125,7 +125,8 @@ public interface SlangCompiler {
      * @return an {@link ExecutableModellingResult} object, containing an executable which is either a flow
      *     or an operations in the file, and a list of all the errors that were found (if any).
      */
-    ExecutableModellingResult preCompileSource(SlangSource source, PrecompileStrategy precompileStrategy, SensitivityLevel sensitivityLevel);
+    ExecutableModellingResult preCompileSource(SlangSource source, PrecompileStrategy precompileStrategy,
+                                               SensitivityLevel sensitivityLevel);
 
     /**
      * Remove all elements in pre-compile cache. No-cached calls are not affected.
--- a/cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/SlangCompiler.java
+++ b/cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/SlangCompiler.java
@@ -125,7 +125,8 @@ public interface SlangCompiler {
      * @return an {@link ExecutableModellingResult} object, containing an executable which is either a flow
      *     or an operations in the file, and a list of all the errors that were found (if any).
      */
-    ExecutableModellingResult preCompileSource(SlangSource source, PrecompileStrategy precompileStrategy,
+    ExecutableModellingResult preCompileSource(SlangSource source,
+                                               PrecompileStrategy precompileStrategy,
                                                SensitivityLevel sensitivityLevel);
 
     /**
--- a/cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/SlangCompiler.java
+++ b/cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/SlangCompiler.java
@@ -14,6 +14,7 @@ import io.cloudslang.lang.compiler.modeller.result.CompilationModellingResult;
 import io.cloudslang.lang.compiler.modeller.result.ExecutableModellingResult;
 import io.cloudslang.lang.compiler.modeller.result.SystemPropertyModellingResult;
 import io.cloudslang.lang.entities.CompilationArtifact;
+import io.cloudslang.lang.entities.SensitivityLevel;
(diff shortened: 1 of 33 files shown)


cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/SlangCompiler.java
```diff
@@ -109,6 +116,19 @@ CompilationModellingResult compileSource(
...
+     * Pre-compile a CloudSlang source into an {@link ExecutableModellingResult}.
+     * All errors that are found during pre-compilation are collected (an exception is not thrown)
+     *
+     * @param source             the {@link SlangSource}
+     * @param precompileStrategy whether to use caching in pre-compile.
+     * @param sensitivityLevel
+     * @return an {@link ExecutableModellingResult} object, containing an executable which is either a flow
+     *     or an operations in the file, and a list of all the errors that were found (if any).
+     */
+    ExecutableModellingResult preCompileSource(SlangSource source, PrecompileStrategy precompileStrategy,
```
Bonczidai: note: if I am using multi-line params, I like each param oin a new line like you did previously:
```java
    CompilationArtifact compile(
            SlangSource source,
            Set<SlangSource> path,
            PrecompileStrategy precompileStrategy,
            SensitivityLevel sensitivityLevel);
```

cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/modeller/transformers/Transformer.java
```diff
@@ -26,6 +27,8 @@
 
     String keyToTransform();
 
+    void setSensitivityLevel(SensitivityLevel sensitivityLevel);
```
Bonczidai: At compiler/modelling level you added overloaded methods (stateless), but at transformer level you added a flag / setter method (stateful) - I don't think you should mix these behaviors

IMHO opinion you shoul overload the transformer too:
```
    TransformModellingResult<T> transform(F rawData);
    TransformModellingResult<T> transform(F rawData, sensitivityLevel);

    delete this method: --------------> void setSensitivityLevel(SensitivityLevel sensitivityLevel);
```

cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/modeller/ExecutableBuilder.java
```diff
@@ -366,7 +370,8 @@ private ActionModellingResult compileAction(Map<String, Object> actionRawData) {
 
         String errorMessagePrefix = "Action syntax is illegal.\n";
         actionData.putAll(
-                transformersHandler.runTransformers(actionRawData, actionTransformers, errors, errorMessagePrefix));
+                transformersHandler.runTransformers(actionRawData, actionTransformers, errors, errorMessagePrefix,
+                        SensitivityLevel.ENCRYPTED));
```
Bonczidai: I see that for executable in/out params you use the param but for step in/params and actions you use hardcoded value? is that the intended behavior?

cloudslang-entities/src/main/java/io/cloudslang/lang/entities/bindings/values/SensitiveValue.java
```diff
@@ -68,7 +79,12 @@ public void encrypt() {
     protected String encrypt(Serializable originalContent) {
         byte[] serialized = serialize(originalContent);
         String serializedAsString = Base64.encodeBase64String(serialized);
-        return EncryptionProvider.get().encrypt(serializedAsString.toCharArray());
+        Encryption encryption = EncryptionProvider.get();
```
gituser173: we don't need to do that as we use obfuscation only in this particular case and the decrypt method handles both deobfuscation and decryption. @lucian-cm WDYT?

cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/modeller/transformers/Transformer.java
gituser173: @lucian-cm WDYT about this as we already discussed this?

cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/modeller/ExecutableBuilder.java
gituser173: you might be right, I will need to do more testing to confirm 👍

cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/modeller/transformers/Transformer.java
lucian-cm: @gituser173 @Bonczidai I think it is better to do it using overload, it is more natural to keep it stateless.

cloudslang-entities/src/main/java/io/cloudslang/lang/entities/bindings/values/SensitiveValue.java
lucian-cm: Nice catch @Bonczidai .
I think indeed the encryptor deobfuscates and decrypts, but I think this is an implementation side effect and we should provide the spi in the best possible way using `deobfuscate` method in the contract with symmetric logic for both encryption and decryption.

gituser173: @lucian-cm @Bonczidai based on what parameter should I make the choice to decrypt or to deobfuscate?
or you would like to duplicate code from the encryptor?

Bonczidai:
@gituser173 is this ready for review? build is not passing and code seems inconsistent - please use the label (in progress / in review) accordingly

please update PR title with descriptive message (~ can be same as updated issue title)

chronical:
@gituser173 i'd like to double on that, please update the issue title accordingly, it will be needed for future tracking

cloudslang-entities/src/test/java/io/cloudslang/lang/entities/encryption/EncryptorConfigTest.java
```diff
@@ -82,6 +82,16 @@ public String encrypt(char[] clearText) {
                 }
 
                 @Override
+                public String obfuscate(String cypherText) {
```
lucian-cm: `String `clearText

cloudslang-spi/src/main/java/io/cloudslang/lang/spi/encryption/Encryption.java
```diff
@@ -25,6 +25,22 @@
...
+     */
+    String obfuscate(String cypherText);
+
+    /**
+     * Deobfuscates an obfuscated string to a clear text char array
+     *
+     * @param cypherText The obfuscated string
+     * @return The clear text char array
+     */
+    char [] deobfuscate(String cypherText);
```
lucian-cm: `String `clearText

cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/modeller/ExecutableBuilder.java
Bonczidai: @gituser173  what is the resolution here? IMHO you should use the dynamic param to make sure senstivity level is uniform

cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/modeller/transformers/BreakTransformer.java
```diff
@@ -41,6 +42,11 @@
     }
 
     @Override
+    public TransformModellingResult<List<String>> transform(List<String> rawData, SensitivityLevel sensitivityLevel) {
```
Bonczidai: IMHO you should not return null here
`1`: in the transformers that are not relevant to the sensitivity level, return the `transform(rawData)` and call the transform uniformly in TransformersHandler
`2`: leave transformers handler as-is right now, and throw NotImplementedException, NotSupportedException here

@lucian-cm WDYT?

cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/modeller/transformers/WorkFlowTransformer.java
```diff
@@ -30,6 +31,12 @@
     }
 
     @Override
+    public TransformModellingResult<Map<String, Object>> transform(Map<String, Object> rawData,
+                                                                   SensitivityLevel sensitivityLevel) {
+        return null;
```
lucian-cm: please add an exception to prevent accidental programming usage of this method

cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/modeller/ExecutableBuilder.java
gituser173: @Bonczidai the param is needed only for the InputsTransformer, we don't need to pass this param for the other transformers

PR #1016, opened 2017-10-13
Compile steps with do_external tag

ligiacentea:
Compile CloudSlang steps with do external steps.
Skip reference validation for external steps.


cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/scorecompiler/ScoreCompilerImpl.java
```diff
@@ -105,7 +105,11 @@ public ExecutionPlan convert(Executable compiledExecutable) {
             Collection<Executable> executables = new ArrayList<>(filteredDependencies.values());
             executables.add(executable);
 
-            executionPlan.setSubflowsUUIDs(new HashSet<>(dependencies.keySet()));
+            HashSet<String> subflowsUuids = new HashSet<>(dependencies.keySet());
+            if (executable.getExternalExecutableDependencies() != null) {
```
victorursan: at this moment there is no way `getExternalExecutableDependencies()` is `null`, you could check if it's empty or at least add it as a extra condition for the if

cloudslang-compiler/src/test/java/io/cloudslang/lang/compiler/ExternalSlangCompilerIntegrationTest.java
```diff
@@ -11,10 +11,14 @@
 
 import io.cloudslang.lang.compiler.modeller.model.ExternalStep;
 import io.cloudslang.lang.compiler.modeller.model.Flow;
+
 import io.cloudslang.lang.compiler.modeller.model.Step;
 import org.hamcrest.Matchers;
 import org.junit.Ignore;
+import java.util.HashSet;
```
victorursan: this import could join the group from lines: 27-30 😆

cloudslang-compiler/src/test/resources/external/flow_with_external_step_by_path.sl
```diff
@@ -0,0 +1,21 @@
...
+#   http://www.apache.org/licenses/LICENSE-2.0
+
+namespace: cloudslang.external
+
+flow:
+  name: flow_with_external_step_by_path
+  workflow:
+    - step1:
+        do_external:
+          test/flow/Random Number Generator:
```
victorursan: I think it was decided the reference for external steps to be in single quotes

cloudslang-compiler/src/test/resources/external/flow_with_external_step_by_uuid.sl
```diff
@@ -0,0 +1,21 @@
...
+#   http://www.apache.org/licenses/LICENSE-2.0
+
+namespace: cloudslang.external
+
+flow:
+  name: flow_with_external_step_by_uuid
+  workflow:
+    - step1:
+        do_external:
+          2c22355a-9c26-4a9e-893c-01a35e4bb0e3:
```
victorursan: same here 😄

cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/validator/CompileValidatorImpl.java
```diff
@@ -67,9 +68,12 @@
         Set<Executable> flowReferences = new HashSet<>();
 
         for (Step step : steps) {
-            Executable reference = dependencies.get(step.getRefId());
-            errors.addAll(validateStepAgainstItsDependency(flow, step, dependencies));
-            flowReferences.add(reference);
+            // validate all steps, except external steps, that are not in the dependencies list
+            if (!(step instanceof ExternalStep)) {
```
victorursan: Maybe `Step` could be an interface and have a method `getType` that returns whether it's `External` or `Regular` (enum). This would avoid the `instanceof` check. (this comment is regarding all instanceof checks of this type)

ligiacentea: I don't actually see what is the advantage of having a Step interface with two implementations and a getType method? I think that having an extra method to determine whether a step is external or regular might be redundant, when we already know that external steps have a different type then regular steps.
However, it is possible, in this situation to have a polymorphic method requiresValidation and return true or false depending on the case.

victorursan: fair point, then as you said, maybe a method `requiresValidation`?

ligiacentea: Will do it!

PR #1022, opened 2017-10-19
Enhancement - builder and CLI assembly packages

tethryus:
Added assembly profile for CLI and builder
Changed assembly packages logic

Signed-off-by: moldovso <sorin@hpe.com>


cloudslang-cli/pom.xml
```diff
@@ -332,8 +338,125 @@
...
+                                    <outputDirectory>${project.build.directory}/cslang-cli</outputDirectory>
+                                    <resources>
+                                        <resource>
+                                            <directory>${project.build.directory}/cs-content</directory>
+                                            <filtering>false</filtering>
+                                            <includes>
+                                                <include>**</include>
+                                            </includes>
+                                            <excludes>
+                                                <exclude>**/ci-env/**</exclude>
```
Bonczidai: wouldn't it be more robust to specify explicitly what to include?

cloudslang-cli/pom.xml
```diff
@@ -332,8 +338,125 @@
...
+                                            </excludes>
+                                        </resource>
+                                    </resources>
+                                </configuration>
+                            </execution>
+                        </executions>
+                    </plugin>
+                    <plugin>
+                        <groupId>org.apache.maven.plugins</groupId>
+                        <artifactId>maven-assembly-plugin</artifactId>
```
Bonczidai: what about the default libs in `/maven/repo/ folder? What will be the decision?

cloudslang-cli/pom.xml
tethryus: Good point. Will add only what to include.

cloudslang-cli/pom.xml
tethryus: It would be nice to have them all downloaded and the content ready when first running the CLI without having to download any other resources.

cloudslang-cli/pom.xml
```diff
@@ -416,16 +416,11 @@
                                             <directory>${project.build.directory}/cs-content</directory>
                                             <filtering>false</filtering>
                                             <includes>
-                                                <include>**</include>
+                                                <include>**/content/**</include>
```
Bonczidai: great 👍

cloudslang-cli/pom.xml
Bonczidai: agree - please add the basic `@action` dependencies - please also investigate the size of CLI zip after these changes

PR #1026, opened 2017-11-16
Added sensitivity level to do transformer

gituser173:


cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/modeller/transformers/WorkFlowTransformer.java
```diff
@@ -27,13 +27,13 @@
 
     @Override
     public TransformModellingResult<Map<String, Object>> transform(Map<String, Object> rawData) {
-        return new BasicTransformModellingResult<>(rawData, new ArrayList<RuntimeException>());
+        return transform(rawData, SensitivityLevel.ENCRYPTED);
```
Bonczidai: since this is the default sensitivity level for the whole system - I think we should extract it somewhere -> in the tranformer interface, or a new service that returns this, constants class etc..

PR #1032, opened 2018-01-16
Add new contexts

victorursan:


cloudslang-runtime/src/main/java/io/cloudslang/lang/runtime/steps/StepExecutionData.java
```diff
@@ -144,6 +148,23 @@ public void endStep(@Param(ScoreLangConstants.RUN_ENV) RunEnvironment runEnv,
         try {
             Context flowContext = runEnv.getStack().popContext();
 
+            final int flowDepth = runEnv.getParentFlowStack().size();
```
gituser173: maybe extract a method for this part of the code

PR #1038, opened 2018-02-26
Added dependency download for cslang-cli-with-content

tethryus:
Signed-off-by: moldovso <sorin@hpe.com>


tethryus:
@Bonczidai There is a small issue with a system dependency from cs-vmware, but there should be a fix for that, soon.

larisabratean:
if changes were not needed for 7 years, I will close the pr

levice14:
@larisabratean that seems fair : ))

PR #1040, opened 2018-04-26
add generic step data consumer

cristian-nitoiu:


cloudslang-runtime/src/main/java/io/cloudslang/lang/runtime/steps/DefaultSlangStepDataConsumer.java
```diff
@@ -0,0 +1,24 @@
+/*******************************************************************************
+ * (c) Copyright 2016 Hewlett-Packard Development Company, L.P.
```
florinmm: The copyright should be to Micro Focus

cristian-nitoiu: I saw no template in the code for Micro Focus copyright, from what I understood all the HP copyright will be eventually bulk changed to Micro Focus anyway

PR #1045, opened 2018-06-14
Fix cloudslang-cli startup on Java 9 version and above. (java.xml.bin…

George-Giloan:
…d module is not resolved by default on classpath and has been marked deprecated - expected to be completely removed)

Signed-off-by: George Giloan <giloan@hpe.com>


tethryus:
👍

tethryus:
🚲

PR #1057, opened 2019-02-21
seq action test

gituser173:


cloudslang-runtime/src/main/java/io/cloudslang/lang/runtime/steps/CloudSlangSequentialExecutionParametersProviderImpl.java
```diff
@@ -14,13 +14,17 @@
...
+import java.io.Serializable;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 
 import static org.apache.commons.lang3.StringUtils.substring;
 
-public class CloudSlangSequentialExecutionParametersProviderImpl implements SequentialExecutionParametersProvider {
+public class CloudSlangSequentialExecutionParametersProviderImpl implements SequentialExecutionParametersProvider,
+        Serializable {
```
George-Giloan: Why make it serializable?

gituser173: it is used for remoting

PR #1061, opened 2019-03-11
Add unit tests for RPA changes

alexandcosma:


cloudslang-compiler/src/test/java/io/cloudslang/lang/compiler/scorecompiler/ExecutionStepFactoryTest.java
```diff
@@ -240,4 +243,20 @@ public void testCreateJoinBranchesStep() throws Exception {
...
+        Map<String, Serializable> actionData = new HashMap<>();
+        HashMap<String, String> seqActionData = new HashMap<>();
+
+        seqActionData.put("key", "value");
+        actionData.put("sequential_action", seqActionData);
+
+        ExecutionStep result;
+        result = factory.createActionStep(1L, actionData);
+
+        assertEquals(result.getActionData().get(ScoreLangConstants.ACTION_TYPE), ActionType.SEQUENTIAL);
```
CimpeanRadu: as a rule of thumb, assertions in junit are structured as follows: assertEquals(expectedValue, actualValue). keep this in mind to avoid confusions when a test fails and the message will be something like Assertion failure: expected [x] but got [y].

PR #1078, opened 2019-05-07
added new constant

cid77:


cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/modeller/ExecutableBuilder.java
```diff
@@ -332,6 +341,15 @@ public ExecutableModellingResult transformToExecutable(ParsedSlang parsedSlang,
         }
     }
 
+    private void validateOutputs(List<Output> outputs) {
+        for (Output output: outputs) {
+            if (output.hasRobotProperty()) {
+                throw new RuntimeException("'robot' property allowed only for outputs of sequential_action. " +
```
Tirla-Alin: Use the constants for robot and sequential_action.

Tirla-Alin: This throws the error at the first invalid output. I would go with with something like:
return outputs.stream().filter(r->r.has..()).map(r->new RTE()).collect(toList()) and all to the errors list.

cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/modeller/transformers/OutputsTransformer.java
```diff
@@ -30,6 +31,12 @@
  */
 public class OutputsTransformer extends AbstractOutputsTransformer implements Transformer<List<Object>, List<Output>> {
 
+    private static final List<String> ACTIVITY_OUTPUTS_KNOWN_KEYS =
+            Arrays.asList(SENSITIVE_KEY, VALUE_KEY, SEQ_OUTPUT_ROBOT_KEY);
+
+    protected List<String> getKnownKeys() {
```
Tirla-Alin: Why do we need this? Can't the SEQ_OUTPUT_ROBOT_KEY be added to the abstract transformer?

cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/modeller/transformers/AbstractOutputsTransformer.java
```diff
@@ -133,12 +134,25 @@ protected Output createPropOutput(Map.Entry<String, Map<String, Serializable>> e
...
+
+        return output;
+    }
+
+    protected List<String> getKnownKeys() {
+        return KNOWN_KEYS;
     }
 
-    private void validateKeys(Map.Entry<String, Map<String, Serializable>> entry, Map<String, Serializable> props) {
+    protected void validateKeys(Map.Entry<String, Map<String, Serializable>> entry, Map<String, Serializable> props) {
```
Tirla-Alin: why is it required to be protected?

cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/modeller/transformers/AbstractOutputsTransformer.java
```diff
@@ -133,12 +134,25 @@ protected Output createPropOutput(Map.Entry<String, Map<String, Serializable>> e
...
+                createOutput(outputName, value, sensitive);
+
+        if (props.containsKey(SEQ_OUTPUT_ROBOT_KEY)) {
+            output.setRobot((boolean) props.get(SEQ_OUTPUT_ROBOT_KEY));
+        }
+
+        return output;
+    }
+
+    protected List<String> getKnownKeys() {
```
Tirla-Alin: looks like we don't need this.

cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/modeller/ExecutableBuilder.java
```diff
@@ -278,6 +281,11 @@ public ExecutableModellingResult transformToExecutable(ParsedSlang parsedSlang,
                 try {
                     systemPropertyDependencies = dependenciesHelper
                             .getSystemPropertiesForOperation(inputs, outputs, results);
+
+                    if (!isSeqAction) {
```
Tirla-Alin: I would add this validation outside of the scope of switch flow/operation

cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/modeller/transformers/OutputsTransformer.java
cid77: Because we don't want to allow this property in the "publish" section of a step. This transformer handles the flow/activity outputs.

PR #1085, opened 2019-05-30
implemented support for UFT Wait step

cid77:


cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/modeller/transformers/SeqStepsTransformer.java
```diff
@@ -50,6 +60,19 @@
...
+    private static final String WAIT = "Wait";
+    private static final String INVALID_WAIT_SYNTAX = "Syntax for 'Wait' command is not valid.";
+
+    /**
+     * Regex for UFT Wait command. It has two forms:
+     * 1) Wait seconds[,millis]
+     * 2) Wait(seconds[,millis])
+     */
+    private static final Pattern WAIT_REGEX = compile(
+            "Wait[ \\t]+\\d+[ \\t]*(,[ \\t]*\\d+)?[ \\t]*" +
```
lucian-cm: e.g. `compile("Wait\\s+\\d+\\s*(,\\s*\\d+)?\\s*", Pattern.UNICODE_CASE | Pattern.MULTILINE);` for form 1 to allow yaml multiline
with > 
if the multiline gets passed to this `transform` method, I have not tested this.
Or another idea is to add `\\R` for newline everywhere in the pattern.

lucian-cm: I would also suggest to validate for treat things like
`Wait 00002, 0005` as not valid since they might cause problems.

Maybe change the \\d+ to be `(0|[1-9]\\d*)`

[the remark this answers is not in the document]
cid77: \s includes new line

PR #1101, opened 2019-08-26
 Cloudslang precondition step

Platonicaaa:
Added canExecute microstep to execution plan.
Added ExecutionPreconditionSerice where you can add any preconditions to be fulfilled before the start of the execution

Signed-off-by: platon <flavius-rares.platon@microfocus.com>

### DIFF
--- a/cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/scorecompiler/ExecutionPlanBuilder.java
+++ b/cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/scorecompiler/ExecutionPlanBuilder.java
@@ -45,7 +45,8 @@ public class ExecutionPlanBuilder {
     private static final int NUMBER_OF_STEP_EXECUTION_STEPS = 2;
     private static final int NUMBER_OF_PARALLEL_LOOP_EXECUTION_STEPS = 2;
     private static final long FLOW_END_STEP_ID = 0L;
-    private static final long FLOW_START_STEP_ID = 1L;
+    private static final long FLOW_PRECONDITION_STEP_ID = 1L;
+    private static final long FLOW_START_STEP_ID = 2L;
 
     public ExecutionPlan createOperationExecutionPlan(Operation compiledOp) {
         ExecutionPlan executionPlan = new ExecutionPlan();
@@ -93,7 +94,8 @@ public class ExecutionPlanBuilder {
         executionPlan.setFlowUuid(compiledFlow.getId());
         executionPlan.setWorkerGroup(compiledFlow.getWorkerGroup());
 
-        executionPlan.setBeginStep(FLOW_START_STEP_ID);
+        executionPlan.setBeginStep(FLOW_PRECONDITION_STEP_ID);
+        executionPlan.addStep(stepFactory.createPreconditionStep(FLOW_PRECONDITION_STEP_ID, compiledFlow.getName()));
         //flow start step
         executionPlan.addStep(stepFactory.createStartStep(FLOW_START_STEP_ID, compiledFlow.getPreExecActionData(),
                 compiledFlow.getInputs(), compiledFlow.getName(), ExecutableType.FLOW));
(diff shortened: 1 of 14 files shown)


cloudslang-runtime/src/main/java/io/cloudslang/lang/runtime/steps/ExecutableExecutionData.java
```diff
@@ -217,6 +224,35 @@ public void finishExecutable(@Param(ScoreLangConstants.RUN_ENV) RunEnvironment r
...
+            if (!StringUtils.isEmpty(runEnv.getExecutionPath().getParentPath())) {
+                //If it is start of a sub flow then the check should not happen
+                runEnv.putNextStepPosition(nextStepId);
+                return;
+            }
+
+            while (!executionPreconditionService.canExecute(valueOf(executionRuntimeServices.getExecutionId()))) {
+                try {
+                    logger.warn("Execution precondition not fulfilled. Waiting for it to be true.");
+                    Thread.sleep(5_000);
```
Tirla-Alin: 5_000L

cloudslang-runtime/src/main/java/io/cloudslang/lang/runtime/steps/ExecutableExecutionData.java
```diff
@@ -217,6 +224,35 @@ public void finishExecutable(@Param(ScoreLangConstants.RUN_ENV) RunEnvironment r
...
+        try {
+            if (!StringUtils.isEmpty(runEnv.getExecutionPath().getParentPath())) {
+                //If it is start of a sub flow then the check should not happen
+                runEnv.putNextStepPosition(nextStepId);
+                return;
+            }
+
+            while (!executionPreconditionService.canExecute(valueOf(executionRuntimeServices.getExecutionId()))) {
+                try {
+                    logger.warn("Execution precondition not fulfilled. Waiting for it to be true.");
```
Tirla-Alin: I don't think anyone would understand this log. Log the execution id and I would let the executionPreconditionService to log this so we can get more details on the exact reason why the execution can not proceed.

cloudslang-runtime/src/main/java/io/cloudslang/lang/runtime/steps/ExecutableExecutionData.java
```diff
@@ -217,6 +224,35 @@ public void finishExecutable(@Param(ScoreLangConstants.RUN_ENV) RunEnvironment r
...
+                    Thread.sleep(5_000);
+                } catch (InterruptedException e) {
+                    logger.error("Thread was interrupted while waiting for execution precondition to be fulfilled.");
+                }
+            }
+
+            // put the next step position for the navigation
+            runEnv.putNextStepPosition(nextStepId);
+        } catch (RuntimeException e) {
+            logger.error("There was an error running the canExecute execution step of: \'" + nodeName +
```
Tirla-Alin: remove \n\t
also make sure you don't log this error 2 times (you log it, you throw it, I assume somewhere you catch and log it again)

cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/scorecompiler/ExecutionPlanBuilder.java
```diff
@@ -204,8 +206,11 @@ private long getCurrentId(Map<String, Long> stepReferences, Deque<Step> steps) {
             }
         }
 
-        if (step == null || !step.isParallelLoop()) {
-            // the reference is not a step or is not a parallel loop step
+        if (step == null) {
+            // the reference is not a step - usually this means this is the first begin step
+            currentId = FLOW_START_STEP_ID + 1L;
```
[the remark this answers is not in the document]
Platonicaaa: Before adding an extra step at the beginning of the execution plan, the currentId formula was the same for both cases (not a step & not parallel step).

cloudslang-runtime/src/main/java/io/cloudslang/lang/runtime/steps/ExecutableExecutionData.java
Platonicaaa: If we only want the ID to be logged then we could just remove this log entirely, since the ID is logged afterwards by the service anyway

cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/scorecompiler/ExecutionPlanBuilder.java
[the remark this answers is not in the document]
Platonicaaa: I lied. It was sth else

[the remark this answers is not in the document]
Platonicaaa: The formula (before) was the same for the number of flow execution steps and the step execution step. We had to split it due to adding another flow execution step

PR #1103, opened 2019-09-13
No license status

alex26toma:


cloudslang-runtime/src/main/java/io/cloudslang/lang/runtime/steps/ExecutableExecutionData.java
```diff
@@ -235,18 +235,20 @@ public void canExecute(@Param(ScoreLangConstants.RUN_ENV) RunEnvironment runEnv,
...
+                logger.warn("Execution precondition not fulfilled. Waiting for it to be true.");
+
+                if (!executionRuntimeServices.getNoLicenseAvailable()) {
+                    executionRuntimeServices.setNoLicenseAvailable();
                 }
+            } else if (executionRuntimeServices.getNoLicenseAvailable()) {
+                executionRuntimeServices.removeNoLicenseAvailable();
             }
 
             runEnv.putNextStepPosition(nextStepId);
```
Platonicaaa: maybe this should be done only when precondition is fulfilled.

PR #1104, opened 2019-09-16
Extended navigation section. Added ability to specify ROI value.

sergey-ok:


cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/utils/SlangSourceUtils.java
```diff
@@ -38,4 +42,45 @@
...
+        } else {
+            List<Map<String, Serializable>> optionList = (List<Map<String, Serializable>>) value;
+            if (!optionList.isEmpty()) {
+                return (String) optionList.get(0).get(NEXT_STEP);
+            }
+            return null;
+        }
+    }
+
+    public static Serializable setNavigationStepName(Serializable value, String newName) {
```
gituser173: maybe rename this one to getNavigationTarget as it is not really a setter

PR #1105, opened 2019-09-17
scheduler inputs fix

adriana-corui:


cloudslang-runtime/src/main/java/io/cloudslang/lang/runtime/steps/ExecutableExecutionData.java
```diff
@@ -81,6 +81,8 @@ public void startExecutable(@Param(ScoreLangConstants.EXECUTABLE_INPUTS_KEY) Lis
 
             if (userInputs != null) {
                 callArguments.putAll(userInputs);
+                userInputs.forEach((inputName, inputValue) ->
+                        executableInputs.add(new Input.InputBuilder(inputName, inputValue).build()));
```
Platonicaaa: I understand this fixes the issue, but wasn't this done in some other place?
I'm just asking to make sure we are not duplicating the user inputs in some scenarios

PR #1108, opened 2019-10-03
Calculate ROI value in Step Execution.

sergey-ok:


cloudslang-runtime/src/main/java/io/cloudslang/lang/runtime/steps/StepExecutionData.java
```diff
@@ -275,5 +286,30 @@ private String computeWorkerGroup(WorkerGroupStatement workerGroup,
         return workerGroupValue.toString();
     }
 
+    private void putStepNavigationOptions(RunEnvironment runEnv, Map<String, NavigationOptions> stepNavigationOptions) {
+        if (stepNavigationOptions != null) {
+            for (NavigationOptions options: stepNavigationOptions.values()) {
```
gituser173: why for loop?

cloudslang-runtime/src/main/java/io/cloudslang/lang/runtime/steps/StepExecutionData.java
```diff
@@ -275,5 +286,30 @@ private String computeWorkerGroup(WorkerGroupStatement workerGroup,
...
+                break;
+            }
+        }
+    }
+
+    private Double getRoiValue(String stepExecutableResult, Map<String, NavigationOptions> stepNavigationOptions) {
+        if (StringUtils.isNotEmpty(stepExecutableResult) && stepNavigationOptions != null) {
+            NavigationOptions navigationOptions = stepNavigationOptions.get(stepExecutableResult);
+            if (navigationOptions != null && navigationOptions.getOptions() != null) {
+                List<Map<String, Serializable>> options = navigationOptions.getOptions();
```
gituser173: why list of map here? why not Map or list of Pairs?

cloudslang-runtime/src/main/java/io/cloudslang/lang/runtime/steps/StepExecutionData.java
sergey-ok: Here I need to fetch first value.
Instead of loop, I can use something like stepNavigationOptions.values().iterator().next().

cloudslang-runtime/src/main/java/io/cloudslang/lang/runtime/steps/StepExecutionData.java
sergey-ok: Will rewrite this code

PR #1129, opened 2019-12-06
Python exec

alexbajzat:
Changes to include external python executions


cloudslang-runtime/src/main/java/io/cloudslang/lang/runtime/bindings/scripts/ScriptExecutor.java
```diff
@@ -30,17 +30,43 @@
...
 
-    public Map<String, Value> executeScript(Set<String> dependencies, String script, Map<String, Value> callArguments) {
+    private Map<String, Value> runExternalPythonAction(Set<String> dependencies, String script,
+                                                       Map<String, Value> callArguments) {
+        Map<String, Serializable> executionResult = externalPytonRuntimeService.exec(dependencies, script,
+                new HashMap<>()).getExecutionResult();
+        Map<String, Value> result = new HashMap<>();
+
+        for (Map.Entry<String, Serializable> entry : executionResult.entrySet()) {
+            result.put(entry.getKey(), ValueFactory.create(entry.getValue(), false));
```
Tirla-Alin: check if input with same name exists save sensitivity

PR #1137, opened 2020-01-17
External python script validation

ivorobiov88:


cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/modeller/transformers/PythonActionTransformer.java
```diff
@@ -90,6 +92,14 @@
         return new BasicTransformModellingResult<>(transformedData, errors);
     }
 
+    private List<String> getInputs(Map<String, Serializable> rawData) {
+        try {
+            return (List<String>) rawData.get(INPUTS_KEY);
```
alexbajzat: What if null ? You can maybe do something like `Optional.of(rawData.get(INPUTS_KEY).orElse(emptyList())` ??

PR #1145, opened 2020-01-30
Added acceptance for optional inputs + tests

alexbajzat:


cloudslang-compiler/pom.xml
```diff
@@ -75,11 +75,20 @@
             <artifactId>guava</artifactId>
         </dependency>
 
+        <dependency>
+            <groupId>io.cloudslang.lang</groupId>
```
alex26toma: <groupId>${project.groupId}</groupId>

PR #1152, opened 2020-02-05
Set default jython true

iancuapahidean:


cloudslang-runtime/src/main/java/io/cloudslang/lang/runtime/bindings/OutputsBinding.java
```diff
@@ -49,7 +49,7 @@
         if (possibleOutputs != null) {
             for (Output output : possibleOutputs) {
                 String outputKey = output.getName();
-                String errorMessagePrefix = "Error binding output: '" + output.getName();
+                String errorMessagePrefix = "Binding output: '" + output.getName() + " failed";
```
alexbajzat: Why do we need this changed ?

iancuapahidean: Because we wanted to make the error message be more human readable. Repeating "Error" to many times didn't seemed such a good option

PR #1153, opened 2020-02-06
Python3 useJython instead of use_jython

EmaMaier:


cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/modeller/transformers/PythonActionTransformer.java
```diff
@@ -78,13 +80,16 @@
                 }
 
                 if (isExternalPythonExecution(rawData)) {
+                    // snake_case -> camelCase
+                    rawData.put(ScoreLangConstants.PYTHON_ACTION_USE_JYTHON_KEY,
+                            rawData.remove(SlangTextualKeys.PYTHON_ACTION_USE_JYTHON_KEY));
                     // validate script
                     String script = (String) rawData.get(PYTHON_ACTION_SCRIPT_KEY);
```
alexbajzat: Since we're using `InterfaceName.constantName` let's use this convention for consistency, so replace `PYTHON_ACTION_SCRIPT_KEY` with `SlangTextualKeys.PYTHON_ACTION_SCRIPT_KEY`

PR #1174, opened 2020-06-02
list iterator fix for designer

mallireddy15:
In case of nested iterators context is not being passed correctly . I changed the way how the context is passed to cloudslang content.

Existing Behaviour : In nested scenario same context is being passed to all the iteraors and causing the nested iterators to run forever.
Modified Behaviour: One context object for each iterator in nested scenario. 

Signed-off-by: Mallikarjuna <mallikarjunareddy.tipparreddy@microfocus.com>


cloudslang-runtime/src/main/java/io/cloudslang/lang/runtime/steps/CloudSlangJavaExecutionParameterProvider.java
```diff
@@ -146,6 +146,10 @@ private void handleStepSessionContextArgument(Map sessionData, List<Object> args
 
     private void handleSessionContextArgument(Map sessionData, String objectClassName, List<Object> args,
                                               String parameterName, ClassLoader classLoader) {
+        // cloudslang list iterator fix
```
lucian-cm, line 149: please don't use assignment to parameter, please use new local variable.

PR #1175, opened 2020-06-04
add robot group key and transformer

larisabratean:


cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/modeller/ExecutableBuilder.java
```diff
@@ -646,6 +647,7 @@ private StepModellingResult compileStep(
         replaceOnFailureReference(postStepData, onFailureStepName);
 
         String workerGroup = (String)stepRawData.get(SlangTextualKeys.WORKER_GROUP);
+        String robotGroup = (String)stepRawData.get(SlangTextualKeys.ROBOT_GROUP);
```
pintea-eugen, line 650: format code

PR #1182, opened 2020-06-29
new cs functions

cid77:


cloudslang-entities/src/main/java/io/cloudslang/lang/entities/constants/Regex.java
```diff
@@ -43,6 +43,14 @@
...
     public static final String GET_REGEX_WITH_DEFAULT = "get\\((.+?),(.+?)\\)";
     public static final String CHECK_EMPTY_REGEX = "check_empty\\((.+?),(.+?)\\)";
+    public static final String CS_APPEND_REGEX = "cs_append\\((.+?),(.+?)\\)";
+    public static final String CS_PREPEND_REGEX = "cs_prepend\\((.+?),(.+?)\\)";
+    public static final String CS_EXTRACT_NUMBER_REGEX = "cs_extract_number\\((.+?)(,(.+?))?\\)";
+    public static final String CS_REPLACE_REGEX = "cs_replace\\((.+?),(.+?),(.+?)(,(.+?))?\\)";
+    public static final String CS_ROUND_REGEX = "cs_round\\((.+?)\\)";
+    public static final String CS_SUBSTRING_REGEX = "cs_substring\\((.+?),(.+?)(,(.+?))?\\)";
+    public static final String CS_TO_UPPER_REGEX = "cs_to_upper\\((.+?)\\)";
+    public static final String CS_TO_LOWER_REGEX = "cs_to_lower\\((.+?)\\)";
```
alexbajzat, line 53: python allows spaces between method name and arguments' parentheses (as well after arguments and colon). e.g: `def myFunction    (a, b)     :`

cloudslang-runtime/src/main/java/io/cloudslang/lang/runtime/services/ScriptsService.java
```diff
@@ -0,0 +1,71 @@
...
+public class ScriptsService {
+
+    private static final Logger logger = Logger.getLogger(ScriptsService.class);
+
+    private Map<ScriptFunction, String> scriptsMap = new HashMap<>();
+
+    @PostConstruct
+    public void init() {
+        List<ScriptFunction> list = Arrays.asList(
+                ScriptFunction.GET,
```
alexbajzat, line 36: This is again related to my first comment. We have `GET` here, but it's also matched individually

cloudslang-runtime/src/main/resources/scripts/cs_append.py
```diff
@@ -0,0 +1,2 @@
+def cs_append(value, to_append):
+  return value + to_append
```
alexbajzat, line 2: This works only for strings. If the user explicitly casts the arguments to int, it's an addition

cloudslang-runtime/src/main/resources/scripts/cs_extract_number.py
```diff
@@ -0,0 +1,12 @@
+def cs_extract_number(str, count = 1):
+  nums = str.split()
```
alexbajzat, line 2: Same as the above. If we expect string here, we should enforce that with a check. Or else, `str.split()` will fail

cloudslang-runtime/src/main/resources/scripts/cs_prepend.py
```diff
@@ -0,0 +1,2 @@
+def cs_prepend(value, to_prepend):
+  return to_prepend + value
```
alexbajzat, line 2: same as for append

cloudslang-runtime/src/main/resources/scripts/cs_replace.py
```diff
@@ -0,0 +1,2 @@
+def cs_replace(str, old_val, new_val, count = -1):
+  return str.replace(old_val, new_val) if count < 0 else str.replace(old_val, new_val, count)
```
alexbajzat, line 2: str should be checked for string

cloudslang-runtime/src/main/resources/scripts/cs_round.py
```diff
@@ -0,0 +1,2 @@
+def cs_round(val):
+  return round(val)
```
alexbajzat, line 2: it goes for all the rest from functions, we should probably check for the types we expect in the function, or else we'll get unpredictable results

cloudslang-entities/src/main/java/io/cloudslang/lang/entities/constants/Regex.java
cid77: The regex should match the function invocation, not definition. I know that the regex is fairly "generous" (that's how the old ones are as well) and I think it's ok. If they are too complex, it may add a performance penalty. But we can discuss.

cloudslang-runtime/src/main/java/io/cloudslang/lang/runtime/services/ScriptsService.java
cid77: All the CS functions implementations are now in files. This service loads the python code of the functions from files.

cloudslang-runtime/src/main/resources/scripts/check_empty.py
```diff
@@ -0,0 +1,2 @@
+def check_empty(value_to_check, default_value=None):
+  return default_value if value_to_check is None else value_to_check
```
lucian-cm, line 2: missing new line at the end

cloudslang-runtime/src/main/resources/scripts/cs_append.py
lucian-cm, line 2: missing new line at the end

cloudslang-runtime/src/main/resources/scripts/cs_extract_number.py
```diff
@@ -0,0 +1,12 @@
...
+  n = count
+  result = None
+  for x in nums:
+    if x.isdigit():
+      if n == 1:
+        result = x
+        break
+      else:
+        n = n - 1
+  return result
```
lucian-cm, line 12: add new line at end

cloudslang-runtime/src/main/resources/scripts/cs_prepend.py
lucian-cm, line 2: add new line at end

cloudslang-runtime/src/main/resources/scripts/cs_replace.py
lucian-cm, line 2: add new line at end

cloudslang-runtime/src/main/resources/scripts/cs_round.py
lucian-cm, line 2: add new line at end

cloudslang-runtime/src/main/resources/scripts/cs_substring.py
```diff
@@ -0,0 +1,2 @@
+def cs_substring(str, start, end = None):
+  return str[start: end]
```
lucian-cm, line 2: add new line at end

cloudslang-runtime/src/main/resources/scripts/cs_to_lower.py
```diff
@@ -0,0 +1,2 @@
+def cs_to_lower(str):
+  return str.lower()
```
lucian-cm, line 2: add new line at end

cloudslang-runtime/src/main/resources/scripts/cs_to_upper.py
```diff
@@ -0,0 +1,2 @@
+def cs_to_upper(str):
+  return str.upper()
```
lucian-cm, line 2: add new line at end

cloudslang-runtime/src/main/resources/scripts/get.py
```diff
@@ -0,0 +1,3 @@
+def get(key, default_value=None):
+  value = globals().get(key)
+  return default_value if value is None else value
```
lucian-cm, line 3: add new line at end

cloudslang-runtime/src/main/resources/scripts/get_system_property.py
```diff
@@ -0,0 +1,4 @@
+def get_sp(key, default_value=None):
+  accessed(key)
+  property_value = sys_prop.get(key)
+  return default_value if property_value is None else property_value
```
lucian-cm, line 4: add new line at end

cloudslang-runtime/src/main/resources/scripts/cs_append.py
lucian-cm: Nice catch @alexbajzat, I believe we need to force value to string.

cloudslang-runtime/src/main/resources/scripts/cs_extract_number.py
lucian-cm, line 2: @alexbajzat if the variable is not a string no need to force it, it is ok to fail, since the user choose this. Previously the meaning of the function was changed.

cloudslang-runtime/src/main/resources/scripts/cs_replace.py
lucian-cm: @alexbajzat this seems legit, if user want to cast to some number of whatever it is fine to fail, we are not planning to previent user to make mistakes.

cloudslang-runtime/src/main/java/io/cloudslang/lang/runtime/services/ScriptsService.java
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    private String loadScript(String script) {
+        String resourceFilePath = "/scripts/" + script + ".py";
+
+        try (InputStream inputStream = this.getClass().getResourceAsStream(resourceFilePath)) {
+            return IOUtils.toString(inputStream, StandardCharsets.UTF_8);
+
+        } catch (IOException e) {
+            logger.error("Unable to read properties from " + resourceFilePath, e);
```
lucian-cm, line 62: We need to throw an exception, not log.

cloudslang-runtime/src/main/resources/scripts/check_empty.py
cid77: there's no need and it breaks the tests

PR #1209, opened 2020-07-30
fixing Oracle case with null in SystemProperties

mikhail-mishustin:
There is a valid case when python expression may return null for Prompt Message. For example Oracle case with empty System Properties. Before fix promptMessage was not changed for null case and it will cause to evaluate expression on each step and failed in case where there is no functional dependency. The idea is to set Prompt Message to EMPTY if expression successfully evaluated and return null.

Signed-off-by: Mikhail Mishustin mikhail.mishustin@microfocus.com


mikhail-mishustin:
Feedback was implemented

PR #1228, opened 2020-09-02
log exception from the op result

cid77:
for java op, remove message from stacktrace; option to disable this behavior

### DIFF
--- a/cloudslang-runtime/src/main/java/io/cloudslang/lang/runtime/steps/ActionExecutionData.java
+++ b/cloudslang-runtime/src/main/java/io/cloudslang/lang/runtime/steps/ActionExecutionData.java
@@ -241,7 +241,7 @@ public class ActionExecutionData extends AbstractExecutionData {
             stacktrace = writer.toString();
         }
 
-        logger.error("Action's stacktrace\n" + stacktrace);
+        logger.error("Java operation encountered an exception:\n" + stacktrace);
     }
 
     protected Map<String, Serializable> createActionContext(Map<String, Value> context) {
--- a/cloudslang-runtime/src/main/java/io/cloudslang/lang/runtime/steps/ActionExecutionData.java
+++ b/cloudslang-runtime/src/main/java/io/cloudslang/lang/runtime/steps/ActionExecutionData.java
@@ -37,7 +37,6 @@ import java.io.Serializable;
 import java.io.StringReader;
 import java.io.StringWriter;
 import java.util.Collection;
-import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
@@ -45,7 +44,6 @@ import java.util.ListIterator;
 import java.util.Map;
 import java.util.Set;
 import java.util.stream.Collectors;
-import java.util.stream.Stream;
 
 import static io.cloudslang.score.api.execution.ExecutionParametersConsts.EXECUTION_RUNTIME_SERVICES;
 import static io.cloudslang.score.api.execution.ExecutionParametersConsts.SEQUENTIAL;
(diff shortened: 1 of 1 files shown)


cloudslang-runtime/src/main/java/io/cloudslang/lang/runtime/steps/ActionExecutionData.java
```diff
@@ -193,9 +205,44 @@ public void doAction(@Param(EXECUTION_RUNTIME_SERVICES) ExecutionRuntimeServices
...
+
+        String stacktrace = (String) returnMap.get(EXCEPTION);
+
+        if (REMOVE_MESSAGE_FROM_LOGGED_EX) {
+            StringWriter writer = new StringWriter();
+            final PrintWriter printWriter = new PrintWriter(writer);
+
+            BufferedReader reader = new BufferedReader(new StringReader(stacktrace));
+
+            reader.lines().forEach(line -> {
```
lucian-cm, line 231: lambdas should be stateless, should not touch variables from ouside like printWriter, consider having a standard for

cloudslang-runtime/src/main/java/io/cloudslang/lang/runtime/steps/ActionExecutionData.java
```diff
@@ -193,9 +205,44 @@ public void doAction(@Param(EXECUTION_RUNTIME_SERVICES) ExecutionRuntimeServices
...
+                int idx = line.indexOf(MARKER);
+                String str = idx == -1 ? line : line.substring(0, idx + MARKER.length());
+
+                printWriter.println(str);
+            });
+
+            printWriter.close();
```
lucian-cm, line 238: should not close printWriter in memory as it does nothing.

cloudslang-runtime/src/main/java/io/cloudslang/lang/runtime/steps/ActionExecutionData.java
```diff
@@ -245,8 +292,14 @@ private String normalizeJavaGav(String gav) {
     private Map<String, Value> prepareAndRunPythonAction(Collection<String> dependencies, String pythonScript,
                                                          Map<String, Value> callArguments, boolean useJython) {
         if (StringUtils.isNotBlank(pythonScript)) {
-            return scriptExecutor.executeScript(
+            final Map<String, Value> returnedMap = scriptExecutor.executeScript(
                     normalizePythonDependencies(dependencies), pythonScript, callArguments, useJython);
+
+            if (returnedMap.containsKey(EXCEPTION)) {
```
lucian-cm, line 298: isNotEmpty check seems better as you can have a null or "" value for a key.

cloudslang-runtime/src/main/java/io/cloudslang/lang/runtime/steps/ActionExecutionData.java
```diff
@@ -245,8 +292,14 @@ private String normalizeJavaGav(String gav) {
...
+                logger.warn("Action's exception: " + returnedMap.get(EXCEPTION));
```
lucian-cm, line 299: I believe we should use error logging level with a better explanation: _"Python operation encountered an exception: "_

cloudslang-runtime/src/main/java/io/cloudslang/lang/runtime/steps/ActionExecutionData.java
cid77: there may be unflushed data. It doesn't harm either.

cloudslang-runtime/src/main/java/io/cloudslang/lang/runtime/steps/ActionExecutionData.java
```diff
@@ -193,9 +205,44 @@ public void doAction(@Param(EXECUTION_RUNTIME_SERVICES) ExecutionRuntimeServices
...
+
+        if (hasException(returnMap)) {
+            logException(returnMap);
+        }
+
         return handleSensitiveValues(returnMap, currentContext);
     }
 
+    private boolean hasException(Map<String, Serializable> returnMap) {
+        return returnMap.containsKey(RETURN_CODE) && !"0".equals(returnMap.get(RETURN_CODE)) &&
```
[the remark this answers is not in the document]
cid77: what's this?

cloudslang-runtime/src/main/java/io/cloudslang/lang/runtime/steps/ActionExecutionData.java
cid77: the value is not a String as the isNotEmpty function expects

cloudslang-runtime/src/main/java/io/cloudslang/lang/runtime/steps/ActionExecutionData.java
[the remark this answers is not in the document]
lucian-cm: exception should be of type String.

lucian-cm (approved):

PR #1257, opened 2020-11-04
Debugger breakpoints modify flow variables

ilasalexandra:
Signed-off-by: Alexandra Ilas alexandra.ilas@microfocus.com

### DIFF
--- a/cloudslang-runtime/src/main/java/io/cloudslang/lang/runtime/steps/ExecutableExecutionData.java
+++ b/cloudslang-runtime/src/main/java/io/cloudslang/lang/runtime/steps/ExecutableExecutionData.java
@@ -127,13 +127,14 @@ public class ExecutableExecutionData extends AbstractExecutionData {
                 }
             }
             if (systemContext.containsKey(ScoreLangConstants.DEBUGGER_FLOW_INPUTS)) {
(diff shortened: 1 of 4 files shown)


cloudslang-runtime/src/main/java/io/cloudslang/lang/runtime/steps/ExecutableExecutionData.java
```diff
@@ -126,14 +126,22 @@ public void startExecutable(@Param(ScoreLangConstants.EXECUTABLE_INPUTS_KEY) Lis
...
             executableInputs = Collections.unmodifiableList(mutableInputList);
+
             //restore what was already prompted and add newly prompted values
             Map<String, Value> promptedValues = runEnv.removePromptedValues();
             promptedValues.putAll(missingInputHandler.applyPromptInputValues(systemContext, executableInputs));
 
             Map<String, Prompt> promptArguments = runEnv.removePromptArguments();
 
-            if (systemContext.containsKey(ScoreLangConstants.DEBUGGER_INPUTS)) {
+            if (systemContext.containsKey(ScoreLangConstants.DEBUGGER_STEP_INPUTS)) {
```
alexbajzat, line 144: We should change this to `DEBUGGER_EXECUTABLE_INPUTS` everywhere we use it, because it's not really a step input

alexbajzat (approved):

PR #1307, opened 2021-04-06
added a flag to switch between jython and python in conf file of CLI & builder. Upated the version of cloudslang

praveenkumar-aa:
-Updated the cloud-slang version to 2.0.0-SNAPSHOT
-Added a flag to switch between external python and jython in CLI and builder.

### DIFF
--- a/cloudslang-cli/pom.xml
+++ b/cloudslang-cli/pom.xml
@@ -7,7 +7,7 @@
     <parent>
         <artifactId>cloudslang</artifactId>
         <groupId>io.cloudslang.lang</groupId>
-        <version>1.0.261-SNAPSHOT</version>
+        <version>2.0.0-SNAPSHOT</version>
     </parent>
 
     <artifactId>cloudslang-cli</artifactId>
--- a/cloudslang-cli/pom.xml
+++ b/cloudslang-cli/pom.xml
@@ -16,8 +16,6 @@
         <cs-content.revision>master</cs-content.revision>
         <cs-content.type>branch</cs-content.type>
         <assembly.tarLongFileMode>posix</assembly.tarLongFileMode>
-        <python>3.8.7.0</python>
-        <python.path>@BASEDIR@/python-lib/python-3.8.7</python.path>
     </properties>
 
     <dependencies>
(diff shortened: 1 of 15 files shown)


cloudslang-cli/pom.xml
```diff
@@ -276,6 +278,15 @@
                                     <outputDirectory>${basedir}/target/cslang-cli/maven</outputDirectory>
                                     <destFileName>maven</destFileName>
                                 </artifactItem>
+                                <artifactItem>
+                                    <groupId>com.microfocus.itom.oo</groupId>
+                                    <artifactId>python</artifactId>
+                                    <version>${python}</version>
+                                    <classifier>win</classifier>
```
sashutosh, line 285: shouldn't this be bin?

cloudslang-cli/src/main/resources/configuration/cslang.properties
```diff
@@ -7,4 +7,5 @@ maven.settings.xml.path=${app.home}/maven/conf/settings.xml
 maven.m2.conf.path=${app.home}/maven/conf/m2.conf
 cloudslang.maven.repo.local=${app.home}/maven/repo
 cloudslang.maven.repo.remote.url=http://repo1.maven.org/maven2
-cloudslang.maven.plugins.remote.url=http://repo1.maven.org/maven2
\ No newline at end of file
+cloudslang.maven.plugins.remote.url=http://repo1.maven.org/maven2
```
nehaaws, line 10: since there is no change in this file , plesae rever to original and do not check-in

nehaaws (approved):

cloudslang-cli/src/main/resources/configuration/cslang.properties
praveenkumar-aa: we are pointing to the external python here

lucian-cm (approved):

cloudslang-cli/pom.xml
praveenkumar-aa: removed the python from here, as we are documenting the steps to setup python.

sashutosh:
run-pr-build

PR #1309, opened 2021-04-14
log4j2 liquibase fix

mallireddy15:


pom.xml
```diff
@@ -211,10 +218,18 @@
                 <version>2.13.3</version>
             </dependency>
 
+<!--            <dependency>-->
```
pintea-eugen, line 221: This dependency can be removed, not just commented out.

pom.xml
```diff
@@ -211,10 +218,18 @@
...
+<!--            </dependency>-->
+
+            <!-- log4j2 dependency for liquibase -->
             <dependency>
-                <groupId>org.slf4j</groupId>
-                <artifactId>slf4j-log4j12</artifactId>
-                <version>1.7.16</version>
+                <groupId>org.apache.logging.log4j</groupId>
+                <artifactId>log4j-slf4j-impl</artifactId>
+                <version>2.14.1</version>
```
pintea-eugen, line 232: I would suggest to stick with the 2.13.3 version, for consistency reasons.

pom.xml
mallireddy15: Removed from pom.xml

pom.xml
mallireddy15: Changed the version as suggested

pintea-eugen (approved):
Looks good! 👍

PR #1382, opened 2022-11-04
Fixed Flaky Test caused by the Map.keySet function

bmk15897:
### Description
Flaky Test found using [NonDex](https://github.com/TestingResearchIllinois/NonDex) by running the command -
`mvn -pl cloudslang-runtime edu.illinois:nondex-maven-plugin:1.1.2:nondex -Dtest=io.cloudslang.lang.runtime.bindings.scripts.ScriptExecutorTest#testExternalPythonValid`

The logged failure for the io.cloudslang.lang.runtime.bindings.scripts.ScriptExecutorTest.testExternalPythonValid -

```
[ERROR] Failures: 
[ERROR]   ScriptExecutorTest.testExternalPythonValid:128 arrays first differed at element [0]; expected:<input[1]> but was:<input[2]>
[INFO] 
[ERROR] Tests run: 1, Failures: 1, Errors: 0, Skipped: 0
```
### Investigation
The test fails at ScriptExecutorTest.java:128 with a comparison error while comparing an expected keyset array and the result from the captured keyset array. The [keySet](https://docs.oracle.com/javase/8/docs/api/java/util/Map.html#keySet--) function of the Map class makes no guarantees as to the iteration order of the keys in the Map object. This makes the test outcome non-deterministic, and the test fails whenever keySet returns a mismatch in order of the keys both the Maps. To fix this, the expected and actual maps' keys should be checked in a more deterministic way so that the assertions do not fail.

### Fix
Expected and actual arrays can be compared without making assumptions about the order of the keys in these Maps. As the [Matchers.containsInAnyOrder](http://hamcrest.org/JavaHamcrest/javadoc/1.3/org/hamcrest/Matchers.html#containsInAnyOrder(org.hamcrest.Matcher...)) function compares the values inside the set of keys without needing order, the test becomes deterministic and ensures that the flakiness from the test is removed.

The PR does not introduce a breaking change.

Bharati Kulkarni <bmk15897@gmail.com>

### DIFF
--- a/cloudslang-runtime/src/test/java/io/cloudslang/lang/runtime/bindings/scripts/ScriptExecutorTest.java
+++ b/cloudslang-runtime/src/test/java/io/cloudslang/lang/runtime/bindings/scripts/ScriptExecutorTest.java
@@ -27,6 +27,7 @@ import io.cloudslang.runtime.impl.python.external.ExternalPythonRuntimeServiceIm
 import io.cloudslang.score.events.EventBus;
 import io.cloudslang.score.events.EventBusImpl;
 import io.cloudslang.utils.PythonScriptGeneratorUtils;
+import org.hamcrest.Matchers;
 import org.junit.Assert;
 import org.junit.Test;
 import org.junit.runner.RunWith;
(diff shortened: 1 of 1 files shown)


bmk15897:
Can you please take a look at this pull request and approve it?

QueueUpX (approved):

QueueUpX:
@bmk15897 This will be merged when a new release is planned.

larisabratean:
will be merged from this pr https://github.com/CloudSlang/cloud-slang/pull/1437 with other refactorings. Thank you for your involvement!

PR #1383, opened 2022-11-10
fix flaky tests

ZhewenFu:
## What's the purpose of this PR

Fix test `testTransformStepWithMissingReqKeys` and `testTransformStepWithEmptyReqKeys`

## Which issue(s) this PR fixes:

Fixes the issue that causes flaky tests. The current test may lead to inconsistent ordering of the ArrayList and thus cause the failure of the tests since the iterator of HashSet is inconsistent. To fix this issue, I utilized TreeSet to make sure the ordering would be consistent when we use the `toString` method.

## Brief changelog

Modified `testTransformStepWithMissingReqKeys` and `testTransformStepWithEmptyReqKeys`
Added `TreeSet` to `SeqStepsTransformer.java`

### DIFF
--- a/cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/modeller/transformers/SeqStepsTransformer.java
+++ b/cloudslang-compiler/src/main/java/io/cloudslang/lang/compiler/modeller/transformers/SeqStepsTransformer.java
@@ -19,6 +19,7 @@ import org.apache.commons.lang.Validate;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashSet;
+import java.util.TreeSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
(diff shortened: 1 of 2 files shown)


ZhewenFu:
@QueueUpX Hi could you help me review this PR and https://github.com/CloudSlang/cloud-slang/pull/1384, or help me connect with people who are responsible for review testing issues? Thank you very much in advance!

QueueUpX (approved):

ZhewenFu:
@QueueUpX Thank you for the review! Could you also help me merge these two PRs into the master since I don't have write access? Also please let me know if there is anything I need to do on my end before merge:)

QueueUpX:
@ZhewenFu we will use pr #1382 since we do not really care about the order, is just the test that is flaky and it will be merged when a new release is planned.

ZhewenFu:
> @ZhewenFu we will use pr #1382 since we do not really care about the order, is just the test that is flaky and it will be merged when a new release is planned.

Hi, I think pr #1382 won't fix this specific test since this test asserts the error message here. To ensure the assert works we need to make the ordering of the ArrayList consistent. Another thing we can do instead of maintaining the ordering is to have multiple assertions for ordering here and passing the test whenever one of the ordering works.I detected the failure of this test through the NonDex tool indicates that the test would fail even if the program works. Let me know if you need further clarification :)

PR #1384, opened 2022-11-11
fix flaky test

ZhewenFu:
## What's the purpose of this PR

Fix test `testMetadataPrettyPrint` and `testPrettyPrint `

## Which issue(s) this PR fixes:

Fixes the issue that causes flaky tests. The current test doesn't guarantee the ordering of `metadataToPrint` due to the inconsistent ordering of `getDeclaredFields` method, and the test would fail in some circumstances. To fix this test, I separate `DESCRIPTION_AND_PREREQUISITES` to `DESCRIPTION ` and `PREREQUISITES `, so we don't need to worry about the ordering.

## Brief changelog

Separated `DESCRIPTION_AND_PREREQUISITES` to `DESCRIPTION ` and `PREREQUISITES `

### DIFF
--- a/cloudslang-cli/src/test/java/io/cloudslang/lang/cli/utils/MetadataHelperTest.java
+++ b/cloudslang-cli/src/test/java/io/cloudslang/lang/cli/utils/MetadataHelperTest.java
@@ -40,10 +40,10 @@ import static org.mockito.Mockito.mock;
 @ContextConfiguration(classes = {MetadataHelperTest.Config.class})
 public class MetadataHelperTest {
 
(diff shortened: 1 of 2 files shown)


QueueUpX:
@ZhewenFu we will use pr #1382 since we do not really care about the order, is just the test that is flaky and it will be merged when a new release is planned.

ZhewenFu:
> @ZhewenFu we will use pr #1382 since we do not really care about the order, is just the test that is flaky and it will be merged when a new release is planned.

Same as the previous PR, pr #1382 is not suitable for this specific test since `DESCRIPTION` and `PREREQUISITES` are two separate fields. Please do let me know if I missed/misunderstood anything here.

PR #1393, opened 2022-12-20
Cases for evalExpr function

opreateodora:


razvanfulea (approved):

cloudslang-runtime/src/main/java/io/cloudslang/lang/runtime/bindings/scripts/ScriptEvaluator.java
```diff
@@ -66,12 +66,9 @@ public Value evalExpr(String expr, Map<String, Value> context, Set<SystemPropert
                           Set<ScriptFunction> functionDependencies) {
         try {
             switch (PYTHON_EVALUATOR) {
```
pentektamas, line 68: Add same logic to testExpr method

PR #1399, opened 2023-01-11
Add pythonServer case for eval

opreateodora:
fix tests


cloudslang-runtime/src/main/java/io/cloudslang/lang/runtime/bindings/scripts/ScriptEvaluator.java
```diff
@@ -121,6 +127,29 @@ private Value doEvaluateExpressionExternalPython(String expr,
         return ValueFactory.create(result.getEvalResult(), getSensitive(pythonContext, accessedResources));
     }
 
+    private Value doEvaluateExpressionServerPython(String expr,
```
adriana-corui, line 130: rename to doEvaluateExpressionExternalPythonExecutor

adriana-corui (approved):

PR #1405, opened 2023-02-15
Check if python executor is alive before attempting to run

QueueUpX:


cloudslang-runtime/src/main/java/io/cloudslang/lang/runtime/bindings/scripts/ScriptEvaluator.java
```diff
@@ -70,6 +71,9 @@ public class ScriptEvaluator extends ScriptProcessor {
     @Autowired
     private ScriptsService scriptsService;
 
+    @Autowired
+    PythonExecutorLifecycleManagerService pythonExecutorLifecycleManagerService;
```
larisabratean, line 75: private

larisabratean (approved):

PR #1422, opened 2023-11-07
Add Ignore to a test which will be fixed very soon

pentektamas:


pintea-eugen (approved):
Ok, if it will be fixed later.

PR #1435, opened 2025-01-23
Fix CS CLI to work with python3

TodorAndrei:
added 3 dependencies in pom.xml, right under "end of score dependencies"

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -182,6 +182,27 @@
             </dependency>
             <!--end of score dependencies-->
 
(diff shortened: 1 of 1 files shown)


a.txt
andreivasiletruta, line 1: @TodorAndrei this file does not seem to be needed, or am I missing something?
cc: @lucian-cm

andreivasiletruta (changes requested):

larisabratean:
no need to merge this changes now after discussing with @TodorAndrei

